Issue #14, opened 2024-08-09
Histogram plot for variables cotaining NaNs


sahahner:
### What happened?

The function `plot_histogram` in `plot.py` does not work for variables containing NaNs. This is for example the case for wave variables, which have NaNs over land.

### What are the steps to reproduce the bug?

Enable the callback for a variable that contains NaNs.
`diagnostics.plot.parameters_histogram`

### Version

-

### Platform (OS and architecture)

atos

PR #15, opened 2024-08-09
fix: histogram only on non-nan values

sahahner:
Fix issue #14


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=15) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=15) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=15) it.</sub>

anaprietonem (approved):
This PR looks to me! As a minor thing I would add a comment to explain that because of the postprocessors being applied, then the data might contains NaNs

sahahner:
In addition to fixing the issue, I noticed that the histogram plot containing histograms for the true and predicted variable does not fix the range of the bins. This leads to bins of different sizes and at different positions. Therefore, the histograms do not allow a fair comparison of the two variables if the values lie in different ranges.
The last commit incorporates a change, which forces the two variables to be binned into the same bins.

The effect on the histogram plotted after few epochs, when the differences are still high, is clearly visible:
Before: with bins covering smaller ranges on the x-axis, densities of true data go up to 3500
image
After: density up to 8000 using the same bins for both 
image

I also changed the name of the label "Truth (ERA5)" to "Truth (data)"

@anaprietonem, can you please have a look at the additional changes?

anaprietonem:
> In addition to fixing the issue, I noticed that ...
So If I am reading correctly what you are doing is passing the range to the histogram so when plotting both the histogram for the predicted and truth data that uses the same range, right?

> I also changed the name of the label "Truth (ERA5)" to "Truth (data)"
About this, so I assume you'd like to remove ERA5 since we might use other training datasets? If so maybe we should change it also in the spectra plot where there is another label like that. What do you think about using 'Predicted' or 'AI-predicted' rather than anemoi?

sahahner:
> So If I am reading correctly what you are doing is passing the range to the histogram so when plotting both the histogram for the predicted and truth data that uses the same range, right?

Yes.

sahahner:
> About this, so I assume you'd like to remove ERA5 since we might use other training datasets? If so maybe we should change it also in the spectra plot where there is another label like that. What do you think about using 'Predicted' or 'AI-predicted' rather than anemoi?

That is a very good proposal. I changed it in the other plot as well.

Yes, ERA5 is not necessarily used by everyone. I also want to point out that this is not the meteorological truth, but what our model considers as true data.

anaprietonem (dismissed):
The changes applied to the histogram function looks good to me!

gmertes (approved):

PR #17, opened 2024-08-12
Changes to data indices in anemoi models

sahahner:
This PR brings the changes in anemoi-models to introduce the remapper into anemoi-training.

- In validation mode, the preprocessor needs to be applied out-of-place instead of in-place since the remapper changes the number of variables. Therefore, in the callbacks the preprocessor needs to be applied.
  - The impact on memory consumption and runtime as been thoroughly tested using the profiler. The results on [mlflow](https://mlflow.copernicus-climate.eu/#/experiments/105?searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D) show that the increase is not significant.
- Incorporate the changes to the data-indices into anemoi-training. After the preprocessor and before the preprocessor we need to use the internal set of indices now.
- Changes to config files
- Cycical colour maps are used to plot variables in degrees.

Some tests will fail because this PR depends on the following PR on anemoi-models:  https://github.com/ecmwf/anemoi-models/pull/21.
When this branch is checked out, all tests pass.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=17) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=17) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=17) it.</sub>

src/anemoi/training/diagnostics/plots.py
```diff
@@ -291,8 +291,8 @@ def plot_histogram(
 
         # Visualization trick for tp
         if variable_name in {"tp", "cp"}:
-            hist_yt *= bins_yt[:-1]
```
sahahner, line 294: `*=` operation not possible for different data types.

src/anemoi/training/config/diagnostics/eval_rollout.yaml
```diff
@@ -71,7 +71,7 @@ log:
     offline: False
     log_model: False
     project: 'Anemoi'
-    entity: ???
+    entity: []
```
mchantry, line 74: Is this an unrelated fix?

gmertes: That's a remnant from when missing values were behaving incorrectly but that's resolved now. Sara, can you change this back to `???`

src/anemoi/training/train/forecaster.py
```diff
@@ -201,18 +218,26 @@ def _step(
     ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor]]:
         del batch_idx
         loss = torch.zeros(1, dtype=batch.dtype, device=self.device, requires_grad=False)
-        batch = self.model.pre_processors(batch)  # normalized in-place
+        # for validation not normalized in-place because remappers cannot be applied in-place
+        batch = (
+            self.model.pre_processors(batch, in_place=False) if validation_mode else self.model.pre_processors(batch)
```
JesperDramsch, line 223: could just be something like `in_place=not validation_mode` instead

src/anemoi/training/train/forecaster.py
```diff
@@ -140,10 +145,10 @@ def metrics_loss_scaling(config: DictConfig, data_indices: IndexCollection) -> t
             pressure_level.minimum,
         )
 
-        for key, idx in data_indices.model.output.name_to_index.items():
+        for key, idx in data_indices.internal_model.output.name_to_index.items():
             # Split pressure levels on "_" separator
             split = key.split("_")
-            if len(split) > 1:
+            if len(split) > 1 and split[1].isdigit():
```
JesperDramsch, line 151: Let's change this to index -1 instead for multi-"_"-variables with pressure levels

src/anemoi/training/diagnostics/plots.py
```diff
@@ -433,42 +433,90 @@ def plot_flat_sample(
...
-            data=truth - input_,
+            data=truth - pred,
             cmap="bwr",
             norm=TwoSlopeNorm(vcenter=0.0),
-            title=f"{vname} persist err",
+            title=f"{vname} pred err",
         )
+
+    if sum(input_) != 0:
+        if vname == "mwd":
```
JesperDramsch, line 474: we should put this on a list to revisit when we refactor the plotting

JesperDramsch: maybe a config setting in the future, but ok for now I think

sahahner: I added an issue under "later"

src/anemoi/training/train/forecaster.py
sahahner: You are right, this is a lot cleaner.

JesperDramsch (approved):
LGTM

PR #18, opened 2024-08-12
Fix null tracker metadata and disable mlflow by default

gmertes:
- This PR fixes the tracker metadata to be empty when there is no tracker, instead of "null".
- Since the tracking URI is now missing by default, I also changed mlflow to be disabled by default, so that the default training config does not crash.

When tracking is disabled, the metadata contains the following for the tracker:

```json
"tracker": {
    "null": null
}
```

Instead, it should be an empty object to indicate there was no tracker:
```json
"tracker": {}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Keep it human-readable, your future self will thank you!
  - Moved callbacks into folder to fascilitate future refactor
  - Adjusted PyPI release infrastructure to common ECMWF workflow
  - Bumped versions in Pre-commit hooks
+ - Fixed "null" tracker metadata when tracking is disabled, now returns an empty dict
 
 ### Removed
  - Dependency on mlflow-export-import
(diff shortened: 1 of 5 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=18) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=18) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=18) it.</sub>

JesperDramsch (dismissed):
Very nice. 

I think we should probably have a check that W&B and MLFlow aren't both enabled, as this would break the tracker info though.

gmertes:
I agree. Since I don't think both trackers should be enabled at the same time, that check could be part of the config validation proposed in #1

PR #19, opened 2024-08-12
Fix hyperparameter logging crash with missing config values

gmertes:
The logger does its own serialisation of the config in `log_hyperparams`, which will trigger missing value errors if any optional config entries are still missing (such as wandb.entity).

This PR makes sure the config entry in the hyper parameters is serialisable without throwing missing value errors. Missing config entries will be resolved to a ??? string.


JesperDramsch (approved):
LGTM thank you

Issue #23, opened 2024-08-13
Plot a validation sample after long rollout


sahahner:
### Is your feature request related to a problem? Please describe.

When wanting to evaluate the stability of the trained model, we might want to have a look at a long rollout for a validation sample.

### Describe the solution you'd like

For the validation sample, for which we already plot the forecast and forecast errors for the rollout steps used during training, plot the forecasts for selected longer rollouts.

### Describe alternatives you've considered

Evaluating longer rollouts is done in inference right now and uses a different tool.

### Organisation

ECMWF

PR #26, opened 2024-08-14
23 plot a validation sample after long rollout

sahahner:
Solves issue #23 

**New callback:** For the validation sample, for which we already plot the forecast and forecast errors for the rollout steps used during training, plot the forecasts for selected longer rollouts using the same layout as the plots we already have.

**Minor additional changes**:

Loss histogram plot:
- Omit the printing of the order of parameters in the loss histogram plot, as this information is included in the legend of the plot thanks to a recent PR.
- Fix: the loss histogram plot broke when using less than 15 variables without proving a parameter group.

forecaster.py:
- Fix: do not consider all variables containing an `_` as pressure level variables, but only those that have a digit indicating the pressure level of the variable. Otherwise, consider as a surface variable.

Possible reviewers: @lzampier


src/anemoi/training/diagnostics/callbacks/__init__.py
lzampier, line 1: You could add `@rank_zero_only` and maybe `@abstractmethod` before the `_plot` in line 299, as done in Line 137 within `BasePlotCallback`.

lzampier (changes requested):
See my suggestions for [CHANGELOG.md](https://github.com/ecmwf/anemoi-training/pull/26/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed) and [src/anemoi/training/diagnostics/callbacks/__init__.py](https://github.com/ecmwf/anemoi-training/pull/26/files#diff-be88bd9789926a500f0714212ab174da8e4d741cf6ef597f23261acbc853dd25) in the specific comments I left.

lzampier (approved):

src/anemoi/training/data/datamodule.py
```diff
@@ -125,6 +125,8 @@ def ds_valid(self) -> NativeGridDataset:
         r = self.rollout
         if self.config.diagnostics.eval.enabled:
             r = max(r, self.config.diagnostics.eval.rollout)
+        if self.config.diagnostics.plot.longrollout.enabled:
```
theissenhelen, line 128: If both if-statements are enabled, the latter will overwrite the value r, is that intentional?

lzampier (approved):

lzampier (approved):

JesperDramsch (approved):
Approving as proxy for @lzampier

PR #30, opened 2024-08-15
Define common paths ignore for downstream CI

JesperDramsch:
This is to limit the runs of downstream ci for pure documentation changes in all triggers using yaml aliases.


theissenhelen (approved):
All good!

PR #38, opened 2024-08-23
Mlflow benchmark profiler update

anaprietonem:
**Describe your changes**
PR to update the benchmark profiler to be able to:

- run and track results with mlflow (including summary of system metrics)
- Add the training and validation rates as metrics monitored by mlflow
- Replace the memory profiler that used memray with torch profiler
- Include option to generate model summary using torchinfo (already a listed dependency)

**Please also include relevant motivation and context.**
Previous version of the BP did not allow to configure the exact reports to be generated and did not make use of torch profiler. In this PR we address those issues, allowing one to choose what reports to generate, exploiting the features available in pytorch profiler for a more in depth breakdown of GPU/CPU memory usage and adding the option to generate a model summary

**List any dependencies that are required for this change.** 
When running the memory report generation, it's possible to perform a Holistic Trace Analysis but for that you'd need to install https://hta.readthedocs.io/en/latest/source/intro/installation.html (so this is now included as extras for the profiler in setup.py)

**Type of change**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] * This change requires a documentation update

**Issue ticket number and link**
We did not create a ticket when we first started working on this task. Will do next time, apologies!

**Checklist before requesting a review**
- [X] I have performed a self-review of my code
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have updated the documentation and docstrings to reflect the changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] I have ensured that the code is still pip-installable after the changes and runs
- [X] I have not introduced new dependencies in the inference partion of the model
- [X] I have ran this on single GPU
- [X] I have ran this on multi-GPU or multi-node
- [ ] I have ran this to work on LUMI (or made sure the changes work independently.)
- [ ] I have ran the Benchmark Profiler against the old version of the code

*need to still update the confluence page but docstrings and comments are updated

Tag possible reviewers
You can @-tag people to review this PR in addition to formal review requests.
@cathalobrien @mchantry @JesperDramsch 

**[PR Migrated from aifs-mono to anemoi-training]**

----
📚 Documentation preview 📚: https://anemoi-training--38.org.readthedocs.build/en/38/


cathalobrien:
Hi, I tested this branch and the various profilers all worked as expected. Nice work!

At first, some of the memory profiler output can be confusing. It shows negative numbers, and positive numbers greater then max device memory. Turns out this is because the profiler tracks deallocations, and aggregates allocations across the entire program runtime. Maybe we could add a note above the memory profiler output explaining this. Or maybe better to link to the documentation for the profiler in stdout, as I understand there's more options which can be enabled if the user digs into the code

| Name | ... | CUDA Mem   | Self CUDA Mem |  # of Calls | ...
| autograd::engine::evaluate_function: AddmmBackward0 | ... | 72.27 Gb   | -44.86 Gb       |          340 | ...

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -4,6 +4,11 @@
 # In applying this licence, ECMWF does not waive the privileges and immunities
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
+
+# * [WHY ARE CALLBACKS UNDER __init__.py?]
+# * This functionality will be restructured in the near future
+# * so for now callbacks are under __init__.py
+
```
theissenhelen, line 11: Harrison has opened a PR for this, so I think we can delete it.

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ Keep it human-readable, your future self will thank you!
 - Feature: Add configurable models [#50](https://github.com/ecmwf/anemoi-training/pulls/50)
 - Feature: Support training for datasets with missing time steps [#48](https://github.com/ecmwf/anemoi-training/pulls/48)
 - Long Rollout Plots
+- Feat: Anemoi Profiler compatible with mlflow and using Pytorch (Kineto) Profiler for memory report
```
mchantry, line 29: Tag the PR please.

docs/user-guide/benchmarking.rst
```diff
@@ -0,0 +1,746 @@
...
+
+The profiler requires certain new packages to be installed, and hence
+has a specific section in the\ ``pyproject.toml``
+(``optional-dependencies.profile``). Hence the first time you'd like to
+use you first need to make sure you have the dependencies installed by
+doing:
+
+.. code:: bash
+
+   pip install -e .[profile]
```
HCookie, line 106: I would recommend this to be a pypi install reference.
```
pip install anemoi-training[profile]
```

src/anemoi/training/commands/profiler.py
```diff
@@ -29,19 +31,55 @@ class Profile(Command):
     def add_arguments(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         return parser
 
-    @staticmethod
-    def run(args: list[str], unknown_args: list[str] | None = None) -> None:
-        del args
+    def run(self, args: list[str], unknown_args: list[str] | None = None) -> None:
+        # This will be picked up by the logger
+        os.environ["ANEMOI_PROFILER_CMD"] = f"{sys.argv[0]} {args.command}"
```
gmertes, line 36: Does this need to be a specific one for the profiler? I think we can just reuse the ANEMOI_TRAINING_CMD env var?

anaprietonem: Yes I wanted to check that! I first opted to have the two of them just to check if it's was working fine, which it does! Right now there is also quite a bit of repeated code across the profiler and train command. So I was thinking I could directly inherit from Train to do the Profiler one to avoid repeating the _merge_sysargv and other functions? the command could even go to a small function so then if I inherit I don't need to code it again. What do you think? (I have not looked  a lot to the details of the Command class, so would like to check thoughts in inheritance could be okey or is not advised in this case)

mchantry (dismissed):
LGTM thanks very much.

pyproject.toml
```diff
@@ -76,17 +76,25 @@ optional-dependencies.docs = [
...
 
 urls.Changelog = "https://github.com/ecmwf/anemoi-training/CHANGELOG.md"
 urls.Documentation = "https://anemoi-training.readthedocs.io/"
 urls.Homepage = "https://github.com/ecmwf/anemoi-training/"
 urls.Issues = "https://github.com/ecmwf/anemoi-training/issues"
 urls.Repository = "https://github.com/ecmwf/anemoi-training/"
+scripts.".anemoi-training-profiler" = "anemoi.training.commands.profiler:main"
+
 # command for interactive DDP (not supposed to be used directly)
 # the dot is intentional, so it doesn't trigger autocomplete
```
gmertes, line 96: Very minor: could you move this comment to be above the profiler entrypoint? Then both entrypoints are below this comment.

src/anemoi/training/commands/profiler.py
```diff
@@ -0,0 +1,55 @@
...
+import logging
+import os
+import sys
+
+from anemoi.training.commands.train import Train
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Profiler(Train):
```
gmertes, line 22: I think inheriting from a common class is a good idea here. But we should avoid inheriting from concrete classes.

Also, like you already saw, you have to define the `run` in the Profiler anyway because it calls the `main()` at the end. If you don't define it, it will call the main from the train command. 

So with both these points in mind, I would suggest the following:

1. In the train command, add a `TrainBase` class that inherits from `Command`. `_merge_sysargv` can go into here.
2. `Train` and `Profiler` become subclasses of `TrainBase`
3. Move lines 28-37 of profiler.py into a new function in `TrainBase`. You can call it something like `prepare_sysargv` 
4. The `run` in `Profiler` and `Train` then become:
```python
def run(self, ...):
    self.prepare_sysargv()
    LOGGER.info(...)
    main()
```

There is still a little duplication because you have to define `run` in both subclasses, but it is minimal.

gmertes (changes requested):
Very nice! Just a minor comment about the class inheritance.

src/anemoi/training/commands/profiler.py
anaprietonem: Thanks for the feedback Gert! as discussed offline this is now implemented. I also reviewed the naming of the command. Right now, the command is 'anemoi-training profiler'. While we could 'rename' it to 'anemoi-training profile' to make it more consistent with 'anemoi-training train'. This is not possible due to conflicts with out pre-commit hooks `"src/anemoi/training/commands/profile.py:1:1: A005 Module `profile` is shadowing a Python builtin module"`  (no file in the codebase can then be called profile.py).
I would then propose to leave it as 'profiler' unless someone else knows a better way to overcome this Ruff

PR #49, opened 2024-09-04
Feat: adding in ability to configure precip like plots

da-ewanp:
**change to precipitation plotting**: Allows you to configure which variables you want to plot with the defined precipitation accumulation levels colourbar and histogram normalisation. This will be useful when adding additional precipitation related fields such as snowfall and runoff in training.

Possible reviewers: @anaprietonem

### DIFF
--- a/src/anemoi/training/config/diagnostics/eval_rollout.yaml
+++ b/src/anemoi/training/config/diagnostics/eval_rollout.yaml
@@ -24,7 +24,7 @@ plot:
   #Defining the accumulation levels for precipitation related fields and the colormap
   accumulation_levels_plot: [0, 0.05, 0.1, 0.25, 0.5, 1, 1.5, 2, 3, 4, 5, 6, 7, 100] # in mm
   cmap_accumulation: ["#ffffff", "#04e9e7", "#019ff4", "#0300f4", "#02fd02", "#01c501", "#008e00", "#fdf802", "#e5bc00", "#fd9500", "#fd0000", "#d40000", "#bc0000", "#f800fd"]
-  precip_and_related_fields: {"tp", "cp"}
+  precip_and_related_fields: [tp, cp]
   # Histogram and Spectrum plots
   parameters_histogram:
   - z_500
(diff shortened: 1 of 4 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=49) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=49) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=49) it.</sub>

src/anemoi/training/diagnostics/plots.py
```diff
@@ -351,6 +358,8 @@ def plot_predicted_multilevel_flat_sample(
         Expected data of shape (lat*lon, nvar*level)
     y_pred : np.ndarray
         Predicted data of shape (lat*lon, nvar*level)
+    precip_like_vars : set, optional
+        Set of precipitation-like variables, by default {"tp", "cp"}
```
anaprietonem, line 362: Do we want to keep this as default? just wondering if it might conflict when someone train using a dataset where there is not tp/cp defined

src/anemoi/training/config/diagnostics/eval_rollout.yaml
```diff
@@ -24,6 +24,7 @@ plot:
   #Defining the accumulation levels for precipitation related fields and the colormap
   accumulation_levels_plot: [0, 0.05, 0.1, 0.25, 0.5, 1, 1.5, 2, 3, 4, 5, 6, 7, 100] # in mm
   cmap_accumulation: ["#ffffff", "#04e9e7", "#019ff4", "#0300f4", "#02fd02", "#01c501", "#008e00", "#fdf802", "#e5bc00", "#fd9500", "#fd0000", "#d40000", "#bc0000", "#f800fd"]
+  precip_like_vars: {"tp", "cp"}
```
anaprietonem, line 27: I am wondering about the naming of this. Is the cmap/features associated to just precip and other associated variables or this due to the bounding/treatment with do on precip?

anaprietonem (commented):
I have left some minor comments!

src/anemoi/training/diagnostics/plots.py
da-ewanp: Good question! If tp/cp aren't defined this default should have no effect. The function just tests if they are in the set {tp, cp} and if not it will use the default plotting. Do you think this is okay?

src/anemoi/training/config/diagnostics/eval_rollout.yaml
da-ewanp: Yes good point, I was thinking we might be able to find a better name. This plotting will only ever really need to be applied to precipitation type variables with units of kg m-2 s-1 I think, so runoff (ro) and snowfall (sf) etc. Not sure if we could go for something like precipitation_related_fields, or moisture_variables, or just leave as precip_like_vars. Do you have any other ideas you think might be clearer?

anaprietonem (approved):
After discussing some of the comments with Ewan offline and having tested the code, I am happy to approve this PR!

PR #56, opened 2024-09-12
Chore/multiple fixes ci precommit

theissenhelen:
fixes/updates:
- correct downstream-ci-hpc workflow call
- ignore pre-commit-config and readthedocs for changelog updates
- updates pre-commit config

adds:

- merge strategy for CHANGELOG.md
- adds CODEOWNERS file

### DIFF
--- a/.github/ci-hpc-config.yml
+++ b/.github/ci-hpc-config.yml
@@ -1,4 +1,5 @@
 build:
+  python: '3.10'
   modules:
     - ninja
   dependencies:
--- a/.github/ci-hpc-config.yml
+++ b/.github/ci-hpc-config.yml
@@ -1,0 +1,18 @@
+build:
+  modules:
+    - ninja
+  dependencies:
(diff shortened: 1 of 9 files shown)


.github/ci-hpc-config.yml
```diff
@@ -0,0 +1,18 @@
+build:
+  modules:
```
anaprietonem, line 2: Probably naive question, but the PR i reviewed for anemoi-models yesterday, in this file it was listed python: '3.10' , we don't we need that here?

anaprietonem (commented):
Left one question just for clarification! after that I will be be happy to approve!

anaprietonem (approved):
I am approving the PR, but as discussed offline we might need to update the hpc-ci config in anemoi-models to also include the pytest command and have a conversation about ways to main the CI/CD folders across so many repos to facilitate the maintenance

PR #58, opened 2024-09-17
Update learning rate schedule in training/default.yaml

jswijnands:
This updates the default values for the learning rate schedule. Specifically, the local learning rate is increased to 8x the original learning rate. In addition, the number of iterations is reduced from 300k to 150k. This has led to similar accuracy in stretched grid experiments using the graph transformer architecture, while requiring only 50% of the GPU compute. Optimised default values could save GPU resources across the Pilot Project.

Suggested review questions:
- Are these settings also an enhancement for global / LAM models? (tested only for stretched grid)
- Are these settings also an enhancement for GNN / transformer models? (tested only for graph transformer architecture)

### DIFF
--- a/src/anemoi/training/config/training/default.yaml
+++ b/src/anemoi/training/config/training/default.yaml
@@ -48,9 +48,9 @@ rollout:
 
 max_epochs: 200
 lr:
-  rate: 0.625e-4 #local_lr
-  iterations: 300000
-  min: 3e-7 #Not scaled by #GPU
+  rate: 5e-4 #local_lr
+  iterations: 150000
+  min: 2.4e-6 #Not scaled by #GPU
 
 # Changes in per-gpu batch_size should come with a rescaling of the local_lr
 # in order to keep a constant global_lr


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=58) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=58) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=58) it.</sub>

JesperDramsch:
Hi @jswijnands, thank you for the contribution.

We'll look into benchmarking this on the non-stretched grid models and evaluate if we should maybe make this a special config, as I believe the stretched-grid may generally benefit from different training defaults.

jswijnands:
Had to create a new fork, which is why this PR was closed automatically. It would be good to still consider these stretched grid defaults as it can save compute and/or time

mchantry:
Thanks Jasper. Yes, we need to test these for other use-cases.

PR #60, opened 2024-09-24
Refactor Callbacks

### LINKED ISSUE #59: Refactor Callbacks
### What maintenance does this project need?

Currently all callbacks are included in one `__init__.py`.
They should be split up into seperate modules and the code to include them refactored.

### Organisation

ECMWF

HCookie:
- Split into seperate files
- Use list in config to add callbacks
- Provide legacy config enabled approach
- Fix ruff issues

## New Usage
Set `config.diagnostics.callbacks` to a list of callback names to include

Closes #59

### DIFF
--- a/src/anemoi/training/diagnostics/callbacks/__init__.py
+++ b/src/anemoi/training/diagnostics/callbacks/__init__.py
@@ -5,1038 +5,55 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-# * [WHY ARE CALLBACKS UNDER __init__.py?]
-# * This functionality will be restructured in the near future
-# * so for now callbacks are under __init__.py
-
 from __future__ import annotations
 
-import copy
 import logging
-import sys
-import time
-import traceback
-import uuid
-from abc import ABC
-from abc import abstractmethod
-from concurrent.futures import ThreadPoolExecutor
-from contextlib import nullcontext
 from datetime import timedelta
-from functools import cached_property
-from pathlib import Path
 from typing import TYPE_CHECKING
-from typing import Any
-from typing import Callable
 
-import matplotlib.patches as mpatches
-import matplotlib.pyplot as plt
-import numpy as np
-import torch
-import torchinfo
-from anemoi.utils.checkpoints import save_metadata
-from pytorch_lightning.callbacks import Callback
-from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
-from pytorch_lightning.utilities import rank_zero_only
-
-from anemoi.training.diagnostics.plots import init_plot_settings
-from anemoi.training.diagnostics.plots import plot_graph_features
-from anemoi.training.diagnostics.plots import plot_histogram
-from anemoi.training.diagnostics.plots import plot_loss
-from anemoi.training.diagnostics.plots import plot_power_spectrum
-from anemoi.training.diagnostics.plots import plot_predicted_multilevel_flat_sample
+from anemoi.training.diagnostics.callbacks.checkpointing import AnemoiCheckpoint
+from anemoi.training.diagnostics.callbacks.evaluation import RolloutEval
+from anemoi.training.diagnostics.callbacks.id import ParentUUIDCallback
+from anemoi.training.diagnostics.callbacks.learning_rate import LearningRateMonitor
+from anemoi.training.diagnostics.callbacks.plotting import GraphTrainableFeaturesPlot
+from anemoi.training.diagnostics.callbacks.plotting import LongRolloutPlots
+from anemoi.training.diagnostics.callbacks.plotting import PlotAdditionalMetrics
+from anemoi.training.diagnostics.callbacks.plotting import PlotLoss
+from anemoi.training.diagnostics.callbacks.plotting import PlotSample
+from anemoi.training.diagnostics.callbacks.weights import StochasticWeightAveraging
 
 if TYPE_CHECKING:
-    import pytorch_lightning as pl
     from omegaconf import DictConfig
-    from omegaconf import OmegaConf
+    from pytorch_lightning.callbacks import Callback
 
 LOGGER = logging.getLogger(__name__)
 
-
-class ParallelExecutor(ThreadPoolExecutor):
-    """Wraps parallel execution and provides accurate information about errors.
-
-    Extends ThreadPoolExecutor to preserve the original traceback and line number.
-
-    Reference: https://stackoverflow.com/questions/19309514/getting-original-line-
-    number-for-exception-in-concurrent-futures/24457608#24457608
-    """
-
-    def submit(self, fn: Any, *args, **kwargs) -> Callable:
-        """Submits the wrapped function instead of `fn`."""
-        return super().submit(self._function_wrapper, fn, *args, **kwargs)
-
-    def _function_wrapper(self, fn: Any, *args: list, **kwargs: dict) -> Callable:
-        """Wraps `fn` in order to preserve the traceback of any kind of."""
-        try:
-            return fn(*args, **kwargs)
-        except Exception as exc:
-            raise sys.exc_info()[0](traceback.format_exc()) from exc
-
-
-class BasePlotCallback(Callback, ABC):
-    """Factory for creating a callback that plots data to Experiment Logging."""
-
-    def __init__(self, config: OmegaConf) -> None:
-        """Initialise the BasePlotCallback abstract base class.
-
-        Parameters
-        ----------
-        config : OmegaConf
-            Config object
-
-        """
-        super().__init__()
-        self.config = config
-        self.save_basedir = config.hardware.paths.plots
-        self.plot_frequency = config.diagnostics.plot.frequency
-        self.post_processors = None
-        self.pre_processors = None
-        self.latlons = None
-        init_plot_settings()
-
-        self.plot = self._plot
-        self._executor = None
-
-        if self.config.diagnostics.plot.asynchronous:
-            self._executor = ParallelExecutor(max_workers=1)
-            self._error: BaseException | None = None
-            self.plot = self._async_plot
-
-    @rank_zero_only
-    def _output_figure(
-        self,
-        logger: pl.loggers.base.LightningLoggerBase,
-        fig: plt.Figure,
-        epoch: int,
-        tag: str = "gnn",
-        exp_log_tag: str = "val_pred_sample",
-    ) -> None:
-        """Figure output: save to file and/or display in notebook."""
-        if self.save_basedir is not None:
-            save_path = Path(
-                self.save_basedir,
-                "plots",
-                f"{tag}_epoch{epoch:03d}.png",
-            )
-
-            save_path.parent.mkdir(parents=True, exist_ok=True)
-            fig.savefig(save_path, dpi=100, bbox_inches="tight")
-            if self.config.diagnostics.log.wandb.enabled:
-                import wandb
-
-                logger.experiment.log({exp_log_tag: wandb.Image(fig)})
-
-            if self.config.diagnostics.log.mlflow.enabled:
-                run_id = logger.run_id
-                logger.experiment.log_artifact(run_id, str(save_path))
-
-        plt.close(fig)  # cleanup
-
-    def teardown(self, trainer: pl.Trainer, pl_module: pl.LightningModule, stage: str) -> None:
-        """Method is called to close the threads."""
-        del trainer, pl_module, stage  # unused
-        if self._executor is not None:
-            self._executor.shutdown(wait=True)
-
-    @abstractmethod
-    @rank_zero_only
-    def _plot(
-        *args: list,
-        **kwargs: dict,
-    ) -> None: ...
-
-    @rank_zero_only
-    def _async_plot(
-        self,
-        trainer: pl.Trainer,
-        *args: list,
-        **kwargs: dict,
-    ) -> None:
-        """To execute the plot function but ensuring we catch any errors."""
-        future = self._executor.submit(
-            self._plot,
-            trainer,
-            *args,
-            **kwargs,
-        )
-        # otherwise the error won't be thrown till the validation epoch is finished
-        try:
-            future.result()
-        except Exception:
-            LOGGER.exception("Critical error occurred in asynchronous plots.")
-            sys.exit(1)
-
-
-class RolloutEval(Callback):
-    """Evaluates the model performance over a (longer) rollout window."""
-
-    def __init__(self, config: OmegaConf) -> None:
-        """Initialize RolloutEval callback.
-
-        Parameters
-        ----------
-        config : dict
-            Dictionary with configuration settings
-
-        """
-        super().__init__()
-
-        LOGGER.debug(
-            "Setting up RolloutEval callback with rollout = %d, frequency = %d ...",
-            config.diagnostics.eval.rollout,
-            config.diagnostics.eval.frequency,
-        )
-        self.rollout = config.diagnostics.eval.rollout
-        self.frequency = config.diagnostics.eval.frequency
-
-    def _eval(
-        self,
-        pl_module: pl.LightningModule,
-        batch: torch.Tensor,
-    ) -> None:
-        loss = torch.zeros(1, dtype=batch.dtype, device=pl_module.device, requires_grad=False)
-        metrics = {}
-
-        # start rollout
-        batch = pl_module.model.pre_processors(batch, in_place=False)
-        x = batch[
-            :,
-            0 : pl_module.multi_step,
-            ...,
-            pl_module.data_indices.internal_data.input.full,
-        ]  # (bs, multi_step, latlon, nvar)
-        assert (
-            batch.shape[1] >= self.rollout + pl_module.multi_step
-        ), "Batch length not sufficient for requested rollout length!"
-
-        with torch.no_grad():
-            for rollout_step in range(self.rollout):
-                y_pred = pl_module(x)  # prediction at rollout step rollout_step, shape = (bs, latlon, nvar)
-                y = batch[
-                    :,
-                    pl_module.multi_step + rollout_step,
-                    ...,
-                    pl_module.data_indices.internal_data.output.full,
-                ]  # target, shape = (bs, latlon, nvar)
-                # y includes the auxiliary variables, so we must leave those out when computing the loss
-                loss += pl_module.loss(y_pred, y)
-
-                x = pl_module.advance_input(x, y_pred, batch, rollout_step)
-
-                metrics_next, _ = pl_module.calculate_val_metrics(y_pred, y, rollout_step)
-                metrics.update(metrics_next)
-
-            # scale loss
-            loss *= 1.0 / self.rollout
-            self._log(pl_module, loss, metrics, batch.shape[0])
-
-    def _log(self, pl_module: pl.LightningModule, loss: torch.Tensor, metrics: dict, bs: int) -> None:
-        pl_module.log(
-            f"val_r{self.rollout}_wmse",
-            loss,
-            on_epoch=True,
-            on_step=True,
-            prog_bar=False,
-            logger=pl_module.logger_enabled,
-            batch_size=bs,
-            sync_dist=False,
-            rank_zero_only=True,
-        )
-        for mname, mvalue in metrics.items():
-            pl_module.log(
-                f"val_r{self.rollout}_" + mname,
-                mvalue,
-                on_epoch=True,
-                on_step=False,
-                prog_bar=False,
-                logger=pl_module.logger_enabled,
-                batch_size=bs,
-                sync_dist=False,
-                rank_zero_only=True,
-            )
-
-    @rank_zero_only
-    def on_validation_batch_end(
-        self,
-        trainer: pl.Trainer,
-        pl_module: pl.LightningModule,
-        outputs: list,
-        batch: torch.Tensor,
-        batch_idx: int,
-    ) -> None:
-        del outputs  # outputs are not used
-        if batch_idx % self.frequency == 0:
-            precision_mapping = {
-                "16-mixed": torch.float16,
-                "bf16-mixed": torch.bfloat16,
-            }
-            prec = trainer.precision
-            dtype = precision_mapping.get(prec)
-            context = torch.autocast(device_type=batch.device.type, dtype=dtype) if dtype is not None else nullcontext()
-
-            with context:
-                self._eval(pl_module, batch)
-
-
-class LongRolloutPlots(BasePlotCallback):
-    """Evaluates the model performance over a (longer) rollout window."""
-
-    def __init__(self, config) -> None:
-        """Initialize RolloutEval callback.
-
-        Parameters
-        ----------
-        config : dict
-            Dictionary with configuration settings
-        """
-        super().__init__(config)
-
-        LOGGER.debug(
-            "Setting up callback for plots with long rollout: rollout = %d, frequency = every %d epoch ...",
-            config.diagnostics.plot.longrollout.rollout,
-            config.diagnostics.plot.longrollout.frequency,
-        )
-        self.rollout = config.diagnostics.plot.longrollout.rollout
-        self.eval_frequency = config.diagnostics.plot.longrollout.frequency
-        self.sample_idx = self.config.diagnostics.plot.sample_idx
-
-    @rank_zero_only
-    def _plot(
-        self,
-        trainer,
-        pl_module: pl.LightningModule,
-        batch: torch.Tensor,
-        batch_idx,
-        epoch,
-    ) -> None:
-
-        start_time = time.time()
-
-        logger = trainer.logger
-
-        # Build dictionary of inidicies and parameters to be plotted
-        plot_parameters_dict = {
-            pl_module.data_indices.model.output.name_to_index[name]: (
-                name,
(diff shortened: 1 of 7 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=60) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=60) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=60) it.</sub>

HCookie:
At the moment, this is the proposed refactor, I am yet to complete an exhaustive test of the changes

JesperDramsch:
Great work, thank you for taking this on.

I was thinking that it might be nice to make this fully configurable through instantiate.

For example, no one is really using the stochastic weight averaging as far as I know, so having specific config entries for this is a bit of feature bloat.

Then the list of callbacks would just look like this:

```
callbacks:
  swa: _target_: pytorch_lightning.callbacks.stochastic_weight_avg.StochasticWeightAveraging
          swa_lr: 1e-4
          swa_epoch_start: 123
          annealing_epochs: 5
          annealing_strategy: cos
          device: null
  blabla: _target_: blabla_callback
             blabla: bla
```

This makes it more extensible and actually reduces some of or less used config entries.

Additionally, we can keep the standard callbacks, like model checkpoints as "permanent callback" (I don't think we have to make everything optional).

One idea I also had is that we could make a special list for "plot_callbacks" in the same style. Then we can easily keep the super convenient "plots.enabled = False" as a shortcut to disable them?

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +86,109 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+def _get_config_enabled_callbacks(config: DictConfig) -> list[Callback]:
+    """Get callbacks that are enabled in the config as according to CONFIG_ENABLED_CALLBACKS
 
-    if config.diagnostics.eval.enabled:
-        trainer_callbacks.append(RolloutEval(config))
+    Provides backwards compatibility
+    """
+    callbacks = []
+
+    def nestedget(conf: DictConfig, key, default):
```
JesperDramsch, line 106: This one needs some comments if we keep it. Took me a second that we're recursively finding a key.

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +86,109 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+        keys = key.split(".")
+        for k in keys:
+            conf = conf.get(k, default)
+            if not isinstance(conf, DictConfig):
+                break
+        return conf
+
+    for enable_key, callback_list in CONFIG_ENABLED_CALLBACKS:
+        if isinstance(enable_key, list):
```
JesperDramsch, line 115: Would it make sense to check for an iterable instead to include tuples?

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +86,109 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+
+    def nestedget(conf: DictConfig, key, default):
+        keys = key.split(".")
+        for k in keys:
+            conf = conf.get(k, default)
+            if not isinstance(conf, DictConfig):
+                break
+        return conf
+
+    for enable_key, callback_list in CONFIG_ENABLED_CALLBACKS:
```
JesperDramsch, line 114: I'm not 100% convinced by this implementation, because I feel this paints us a bit into a corner.

src/anemoi/training/config/diagnostics/eval_rollout.yaml
```diff
@@ -70,6 +70,12 @@ checkpoint:
     save_frequency: null # Does not scale with rollout
     num_models_saved: 0
 
+callbacks:
```
JesperDramsch, line 73: I think it would be good to already populate these

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +86,109 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+        if isinstance(enable_key, list):
+            if not any(nestedget(config, key, False) for key in enable_key):
+                continue
+        elif not nestedget(config, enable_key, False):
+            continue
+
+        if isinstance(callback_list, list):
```
JesperDramsch, line 121: I think after the next iteration of the refactor, this part may be obsolete as we don't have lists in the structure above.

JesperDramsch (changes requested):
Hi @HCookie, thanks for taking on the callbacks!

It's already much better, great work on that. I think we can take the refactor even further and make the callbacks (almost?) fully modular, which would be incredible for future extensibility.

One comment regarding the file names. So far we haven't been using `<xyz>-ing.py` as language. Especially "checkpointing" would be confusing with activation checkpointing (although that is and will stay confusing honestly). Can we rename these please?

src/anemoi/training/diagnostics/callbacks/plotting.py
```diff
@@ -0,0 +1,737 @@
+# ruff: noqa: ANN001
```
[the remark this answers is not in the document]
HCookie: Missing type annotations throughout the priorly made Callbacks

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +86,109 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
     if config.diagnostics.plot.enabled:
-        trainer_callbacks.extend(
-            [
-                PlotLoss(config),
-                PlotSample(config),
-            ],
-        )
         if (config.diagnostics.plot.parameters_histogram or config.diagnostics.plot.parameters_spectrum) is not None:
-            trainer_callbacks.extend([PlotAdditionalMetrics(config)])
+            callbacks.append(plotting.PlotAdditionalMetrics(config))
```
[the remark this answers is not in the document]
HCookie: This is an or operations on included data, rather then the above which were boolean flags

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -5,1038 +5,44 @@
...
-                epoch=epoch,
-                tag=f"gnn_pred_val_sample_rstep{rollout_step:02d}_batch{batch_idx:04d}_rank0",
-                exp_log_tag=f"val_pred_sample_rstep{rollout_step:02d}_rank{local_rank:01d}",
-            )
+# Callbacks to add according to flags in the config
+CONFIG_ENABLED_CALLBACKS: list[tuple[list[str] | str, list[type[Callback]] | type[Callback]]] = [
+    (["diagnostics.log.wandb.enabled", "diagnostics.log.mlflow.enabled"], LearningRateMonitor),
+    ("diagnostics.eval.enabled", RolloutEval),
+    (
+        "diagnostics.plot.enabled",
```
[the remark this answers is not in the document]
HCookie: What do you mean initialise them? Remove and include in the config instead?

src/anemoi/training/diagnostics/callbacks/__init__.py
HCookie: Fair enough, its a little annoying its not in the `DictConfig` by defaut.

src/anemoi/training/diagnostics/callbacks/__init__.py
HCookie: Sure, I suppose considering this a fixed global var it doesn't need to do be too general, but I'll adjust for best practice reasons.

src/anemoi/training/diagnostics/callbacks/__init__.py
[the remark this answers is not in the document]
HCookie: I shall extend the config enabled callbacks to take a Callable to find enabled status

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +144,110 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+    """
+
+    trainer_callbacks: list[Callback] = []
+
+    # Get Checkpoint callback
+    checkpoint_callback = _get_checkpoint_callback(config)
+    if checkpoint_callback is not None:
+        trainer_callbacks.extend(checkpoint_callback)
+
+    # Base callbacks
```
anaprietonem, line 233: Apologies since I might just be slow, but based on this we are mixing different ways of instantiating the callbacks in the config right? Cause we have plotting callbacks and rollout where we use the target method, but then that's not the case for checkpoint and the ones about learning rate? My concern here is when extending to new callbacks it might difficult to decide which way to go, and probably not very intuitive for the user. Could we consider using the instantiation method for all callbacks to keep it consistent?

anaprietonem (commented):
I have left some comments! but great to see you're working on this @HCookie, thanks a lot for the work!

src/anemoi/training/diagnostics/plots.py
```diff
@@ -625,36 +628,163 @@ def scatter_plot(
...
+    -------
+    torch.Tensor of shape (N, 2)
+        Lat/lon coordinates.
+    """
+    ndim = sincos_coords.shape[1] // 2
+    sin_y, cos_y = sincos_coords[:, :ndim], sincos_coords[:, ndim:]
+    return torch.atan2(sin_y, cos_y)
+
+
+def plot_graph_node_features(model: torch.nn.Module) -> Figure:
```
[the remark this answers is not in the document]
JPXKQX: The one for the edges did not exist before. It has been split into 2 callbacks but could be easily merged. It probably makes sense.

JPXKQX:
Thanks for taking the time to refactor this, Harrison! I have two main points that I would like your opinion on:
- What do you think about moving the fields in the config as parameters, sample_idx, ... to each callback. For example, having parameters_histogram under the PlotAdditionalMetrics instance:
```
-  __target__ : anemoi.training.diagnostics.callbacks.plot.PlotAdditionalMetrics
   parameters_histogram: [var1, var2, ...]
   precip_and_related_fields: [tp, cp]
   sample_idx: ${diagnostics.plot.sample_idx}
```

- Should we split PlotAdditionalMetrics into PlotHistograms, PlotSpectrums? For example, I usually set an empty list for parameters_spectrum so that I only get the histograms without the spectrum plots because they are more expensive (computationally).

src/anemoi/training/diagnostics/callbacks/swa.py
```diff
@@ -0,0 +1,63 @@
...
+from pytorch_lightning.callbacks.stochastic_weight_avg import StochasticWeightAveraging as pl_StochasticWeightAveraging
+
+if TYPE_CHECKING:
+    from omegaconf import OmegaConf
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class StochasticWeightAveraging(pl_StochasticWeightAveraging):
```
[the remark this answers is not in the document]
HCookie: I like the idea of an optimiser callback module

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -1081,58 +144,110 @@ def get_callbacks(config: DictConfig) -> list:  # noqa: C901
...
+        if isinstance(key, Callable):
+            return key(config)
+        elif isinstance(key, str):
+            return nestedget(config, key, False)
+        elif isinstance(key, Iterable):
+            return all(nestedget(config, k, False) for k in key)
+        return nestedget(config, key, False)
+
+    for deprecated_key, callback_list in DEPRECATED_CONFIGS:
+        if check_key(config, deprecated_key):
```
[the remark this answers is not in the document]
HCookie: Yes, the DEPRECATED_CONFIGS is to provide compatibility for existing configurations, but warn to suggest a switch over

src/anemoi/training/diagnostics/callbacks/__init__.py
HCookie: I think it definitetly requires some discussion, 
- Some callbacks must be included based on configuration options, see swa, or UUID
- Others make sense to include by default, with a flag
- Others again, are up to the user to include.

The lines to be drawn are tricky, but I would suggest only a very core callback is added via the config enabled options, and the rest the _target_ approach

HCookie: I am leaning to agree that all but the essential should be done throught the _target_ method, including plots

HCookie:
Changes requested have been encorporated, 
- Backwards compatability has been removed
- Callbacks refactored to be more explict with the configuration
- AddtionalPlots have been split

JPXKQX (commented):
Thanks for the refactor, I like the way the config looks now. I have added some comments to discuss.

src/anemoi/training/config/diagnostics/plot/detailed.yaml
```diff
@@ -0,0 +1,65 @@
...
+    epoch_frequency: 5
+  - _target_: anemoi.training.diagnostics.callbacks.plot.PlotLoss
+    # group parameters by categories when visualizing contributions to the loss
+    # one-parameter groups are possible to highlight individual parameters
+    parameter_groups:
+      moisture: [tp, cp, tcw]
+      sfc_wind: [10u, 10v]
+  - _target_: anemoi.training.diagnostics.callbacks.plot.PlotSample
+    sample_idx: 0
+    per_sample : 6
```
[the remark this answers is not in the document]
HCookie: I'll adjust the configs to share a `sample_idx`, it could be changed by the user, but that will be an intentional choice.

src/anemoi/training/config/diagnostics/plot/detailed.yaml
```diff
@@ -0,0 +1,65 @@
+enabled: True # Enable plotting callbacks
```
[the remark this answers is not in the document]
HCookie: Yeah, good point, carried it across from the old code

src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -5,1038 +5,63 @@
...
+    for k in keys:
+        conf = conf.get(k, default)
+        if not isinstance(conf, (dict, DictConfig)):
+            break
+    return conf
+
+
+# Callbacks to add according to flags in the config
+# Can be function to check status from config
+CONFIG_ENABLED_CALLBACKS: list[tuple[list[str] | str | Callable[[DictConfig], bool], type[Callback]]] = [
```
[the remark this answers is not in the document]
HCookie: I had thought that as they are needed depending on other parts of the config it made sense to have them enabled via this method. Therefore, a user won't encounter a weird error where SWA doesn't work when enabled cause they hadn't set the callback.

src/anemoi/training/diagnostics/callbacks/evaluation.py
```diff
@@ -0,0 +1,134 @@
...
+    import pytorch_lightning as pl
+    from omegaconf import OmegaConf
+
+LOGGER = logging.getLogger(__name__)
+
+
+class RolloutEval(Callback):
+    """Evaluates the model performance over a (longer) rollout window."""
+
+    def __init__(self, config: OmegaConf, rollout: int, frequency: int) -> None:
```
HCookie, line 28: Type hint for rollout should be `int | list[int]`

src/anemoi/training/config/diagnostics/evaluation.yaml
```diff
@@ -1,53 +1,8 @@
...
-  # one-parameter groups are possible to highlight individual parameters
-  parameter_groups:
-    moisture: [tp, cp, tcw]
-    sfc_wind: [10u, 10v]
-  learned_features: False
-  longrollout:
-    enabled: False
-    rollout: [60]
-    frequency: 20 # every X epochs
+defaults:
```
mc4117, line 2: This is really neat as I always get confused with all those details. 
I am wondering if plot should be a sub-category of callbacks though? Presumably plot won't work unless callbacks are enabled

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,970 @@
...
+            self._executor.shutdown(wait=True)
+
+    def apply_output_mask(self, pl_module: pl.LightningModule, data: torch.Tensor) -> torch.Tensor:
+        if hasattr(pl_module, "output_mask") and pl_module.output_mask is not None:
+            # Fill with NaNs values where the mask is False
+            data[:, :, ~pl_module.output_mask, :] = np.nan
+        return data
+
+    @abstractmethod
+    @rank_zero_only
```
mc4117, line 142: I know these rank-zeros break if you do ensemble sharding in the ens-score branch. Have you tried running these callbacks with model sharding? From memory it breaks more often in interactive mode

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,970 @@
...
+            torch.cat(tuple(x[self.sample_idx : self.sample_idx + 1, ...].cpu() for x in outputs[1])),
+            in_place=False,
+        )
+        output_tensor = pl_module.output_mask.apply(output_tensor, dim=2, fill_value=np.nan).numpy()
+        data[1:, ...] = pl_module.output_mask.apply(data[1:, ...], dim=2, fill_value=np.nan)
+        data = data.numpy()
+        return data, output_tensor
+
+
+class PlotSpectrum(BasePlotAdditionalMetrics):
```
mc4117, line 804: From the documentation am I right in understanding this only works for tp? I think theoretically it should work for other metrics too

src/anemoi/training/train/forecaster.py
```diff
@@ -241,29 +265,51 @@ def _step(
...
                     y_pred,
                     y,
                     rollout_step,
-                    enable_plot=self.enable_plot,
                 )
-                metrics.update(metrics_next)
-                y_preds.extend(y_preds_next)
+            yield loss, metrics_next, y_pred
+
+    def _step(
```
mc4117, line 293: I'm a bit confused by the difference between rollout_step and step. Would you mind explaining the reasoning for this?

src/anemoi/training/config/diagnostics/evaluation.yaml
HCookie: They exist independently, and there is no enable flag anymore

src/anemoi/training/diagnostics/callbacks/plot.py
HCookie: It would work for any params given in `parameters`, see line 829

src/anemoi/training/diagnostics/callbacks/plot.py
mc4117: I think if you set num_gpus_per_model=2 it should just work https://github.com/ecmwf/anemoi-training/blob/fxi/refactor_callbacks/src/anemoi/training/config/hardware/example.yaml#L10

src/anemoi/training/train/forecaster.py
HCookie: `step` is the simple runner to use the config and calculate the metrics and losses, `rollout_step` is a generator which iteratively rolls the model out allowing for loss calculation. 
They exist separately to allow callbacks to use the rollout function for evaluation. 

step wraps around the rollout and encomposes what a step of the model is, and rollout has a time dim associated

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,968 @@
...
+            diagnostics = [] if self.config.data.diagnostic is None else self.config.data.diagnostic
+            plot_parameters_dict_histogram = {
+                pl_module.data_indices.model.output.name_to_index[name]: (
+                    name,
+                    name not in diagnostics,
+                )
+                for name in self.parameters
+            }
+
+            fig = plot_histogram(
```
JPXKQX, line 871: This should be plot_spectrum

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ Keep it human-readable, your future self will thank you!
 - Feature: New `Boolean1DMask` class. Enables rollout training for limited area models. [#79](https://github.com/ecmwf/anemoi-training/pulls/79)
 
 ### Fixed
+- Refactored callbacks. [#60](https://github.com/ecmwf/anemoi-training/pulls/60)
```
JPXKQX, line 23: This shouldn't be in the 0.2.1 release section. I believe it's related to the merge from main back into develop.

src/anemoi/training/diagnostics/callbacks/plot.py
HCookie: Tested by @JPXKQX, and works

JPXKQX (dismissed):
A long-awaited refactor. Thanks Harrison for the amazing work! I have tested with 1 GPU and 2 GPU (with num_gpus_per_model=2), and all plots in "detailed" option are produced as expected.

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,962 @@
...
+            precision_mapping = {
+                "16-mixed": torch.float16,
+                "bf16-mixed": torch.bfloat16,
+            }
+            prec = trainer.precision
+            dtype = precision_mapping.get(prec)
+            context = torch.autocast(device_type=batch.device.type, dtype=dtype) if dtype is not None else nullcontext()
+
+            with context:
+                self.plot(trainer, pl_module, output, batch, batch_idx)
```
sahahner, line 400: There seems to be a missing argument

```suggestion
                self.plot(trainer, pl_module, output, batch, batch_idx, trainer.current_epoch)
```

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,962 @@
...
+            # Copy to be used across all the training cycle
+            self.post_processors = copy.deepcopy(pl_module.model.post_processors).cpu()
+        if self.latlons is None:
+            self.latlons = np.rad2deg(pl_module.latlons_data.clone().cpu().numpy())
+        local_rank = pl_module.local_rank
+
+        assert batch.shape[1] >= self.rollout + pl_module.multi_step, (
+            "Batch length not sufficient for requested validation rollout length! "
+            f"Set `dataloader.validation_rollout` to at least {self.rollout + pl_module.multi_step}"
+        )
```
sahahner, line 328: self.rollout is in fact a list. Therefore, we need to use max(self.rollout) here.
```suggestion
        assert batch.shape[1] >= max(self.rollout) + pl_module.multi_step, (
            "Batch length not sufficient for requested validation rollout length! "
            f"Set `dataloader.validation_rollout` to at least {max(self.rollout) + pl_module.multi_step}"
        )
```

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -0,0 +1,962 @@
...
+                    data_rollout_step = self.post_processors(input_tensor_rollout_step).numpy()
+
+                    # prepare predicted output tensor for plotting
+                    output_tensor = self.post_processors(
+                        y_pred[self.sample_idx : self.sample_idx + 1, ...].cpu(),
+                    ).numpy()
+
+                    fig = plot_predicted_multilevel_flat_sample(
+                        plot_parameters_dict,
+                        self.config.diagnostics.plot.per_sample,
```
sahahner, line 360: This configuration entry does not exist anymore.

src/anemoi/training/config/diagnostics/plot/detailed.yaml
```diff
@@ -0,0 +1,68 @@
...
+  - _target_: anemoi.training.diagnostics.callbacks.plot.PlotHistogram
+    sample_idx: ${diagnostics.plot.sample_idx}
+    precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+    parameters:
+    - z_500
+    - tp
+    - 2t
+    - 10u
+    - 10v
+  - _target_:  anemoi.training.diagnostics.callbacks.plot.LongRolloutPlots
```
sahahner, line 63: As long as `dataloader.validation_rollout=1`, which is the default, this callback only increases the runtime without providing any additional plots. Should we move it into `rollout_eval.yaml`?

HCookie: We could provide a rollout plots configuration?

sahahner:
In general, this looks good to me. The new layout of the config files is intuitive.
There is one regard I have: The configuration of the callbacks is not traceable via MLFlow, as the list of targets is cut after a certain number of characters in the mlflow parameters.
Is there a way to work around this?

HCookie:
> The configuration of the callbacks is not traceable via MLFlow, as the list of targets is cut after a certain number of characters in the mlflow parameters.

That issue with mlflow is addressed in #91. So once that is merged, the config will be accessible in a dump or fully expanded

JPXKQX (approved):

JPXKQX (dismissed):

sahahner (approved):
Thank you for incorporating the requested changes. This looks good to me now.

PR #70, opened 2024-10-02
Feature/improve loss functions

### LINKED ISSUE #41: Losses / Validation Metrics Implementation
### Issues with Loss Implementation in `anemoi-training`

1. **Nomenclature:**
   - The current implementation uses `self.loss_weights` for weighting in the spatial dimension and `self.data_variances` for weighting in the feature dimension. However, this naming can be confusing given that the default strategy for feature dimension weighting is not based on `data_variances`.
   - **Proposal:** Rename `self.loss_weights` to `self.node_weight` (for spatial dimension) and `self.data_variances` to `self.feature_weight` (for feature dimension). Additionally, modify the function responsible for generating feature scaling values to support various feature scaling strategies.

2. **Consistency in Feature Scaling for Loss Calculation:**
   - The loss calculated by `self.metrics` does not currently apply feature scaling, whereas the loss used during training does apply the prescribed feature scaling. Previously, feature scaling was applied consistently across both metrics and training losses.
   - **Proposal:** Either revert to the previous behavior where feature scaling is applied to both, or introduce an option for users to specify whether feature scaling should be applied in `self.metrics`.

3. **Monitoring and Recording Loss Types During Validation:**
   - Currently, only `WeightedMSELoss` is available for monitoring and recording during validation. This limitation restricts flexibility in evaluating model performance.
   - **Proposal:** Extend the configuration options to allow for:
     a) Selection of the type of loss/metric to record.
     b) Choice of whether to record metrics on processed or unprocessed features when evaluating single features (This is crucial for evaluating performance of model's that use different weighting strategies)

---

NA - Inspect code

### Version

develop

NA

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

ECMWF

HCookie:
# Improve loss functions.
Prior, `WeightedMSELoss` was the hard coded loss for training and metrics. 
This PR changes the loss function configuration to be fully modular, so that any `nn.Module` can be used.

Additionally, this adds the following additonal loss functions into `anemoi-training`
- WeightedMAE
- WeightedRMSE
- WeightedLogCosh
- WeightedHuber
- CombinedLoss

Closes #41

## Arbitary Scalars
See #96 for more information
- Allows for arbitary scalars to be applied to the loss values

----
📚 Documentation preview 📚: https://anemoi-training--70.org.readthedocs.build/en/70/


src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
         torch.Tensor
             Weighted MSE loss
-
         """
+        if pred.ndim == 4:
+            pred = pred.mean(dim=1)
+
+        torch.save(self.node_weights, "node_weights.pt")
+        torch.save(pred, "pred.pt")
+        torch.save(target, "target.pt")
```
JesperDramsch, line 83: Leftover debugging code?

src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
+        feature_scale:
+            If True, scale the loss by the feature_weights
 
         Returns
         -------
         torch.Tensor
             Weighted MSE loss
-
         """
+        if pred.ndim == 4:
```
JesperDramsch, line 78: This needs a comment. What is the intention?

src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
+        torch.save(pred, "pred.pt")
+        torch.save(target, "target.pt")
+
         out = torch.square(pred - target)
 
-        # Use variances if available
-        if hasattr(self, "ivar"):
-            out *= self.ivar
+        if feature_scale and hasattr(self, "feature_weights"):
+            out = (
```
JesperDramsch, line 88: What's the performance implication of this change?

The operation before was in-place, this changes the space and time complexity.

src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
+            out /= self.sum_function(self.node_weights.expand_as(out))
             return self.sum_function(out)
 
-        # Weight by area
-        out *= self.weights[..., None].expand_as(out)
+        # Weight by area, due to weighting construction is analagous to a mean
+        out *= self.node_weights[..., None].expand_as(out)
         # keep last dimension (variables) when summing weights
-        out /= self.sum_function(self.weights[..., None].expand_as(out), axis=(0, 1, 2))
+        out /= self.sum_function(self.node_weights[..., None].expand_as(out))
```
JesperDramsch, line 105: Why do we remove the axis here?

src/anemoi/training/train/forecaster.py
```diff
@@ -127,6 +128,23 @@ def __init__(
...
+
+        Will include additional kwargs if specified in the config.
+
+        E.g.
+            If `include_node_weights: True` is set in the config, and node_weights in kwargs
+             `node_weights` will be included in the config to instantiate the module with.
+        """
+        config = dict(config)
+        for key in kwargs:
+            if config.pop(f"include_{key}", False):
```
JesperDramsch, line 144: I don't think we should start the pattern of modifying config entries, this will get unwieldy very quickly.

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,10 +33,34 @@ swa:
...
+
+  # loss function for metric calculation
+  metrics:
+    # loss class to initialise, can be anything subclassing torch.nn.Module
+    _target_: anemoi.training.losses.mse.WeightedMSELoss
+    # what to include in the loss class initialisation
+    include_node_weights: True
+    include_feature_weights: False
+    # other kwargs
+    ignore_nans: True
```
JesperDramsch, line 63: I have a suspicion that making this configurable will end in more bugs than utility.

E.g. The metrics have to ignore Nans, because they're outside of the imputer. But ideally, a user on the config-side shouldn't have to know about this, hence why we hard-code it.

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,10 +33,34 @@ swa:
...
+    # loss class to initialise, can be anything subclassing torch.nn.Module
+    _target_: anemoi.training.losses.mse.WeightedMSELoss
+    # what to include in the loss class initialisation
+    include_node_weights: True
+    include_feature_weights: True
+    # other kwargs
+    ignore_nans: False
+
+  # loss function for metric calculation
+  metrics:
```
JesperDramsch, line 56: I think it would be great to make metrics a list, so users can register multiple evaluation metrics.

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,10 +33,34 @@ swa:
...
+  # don't enable this by default until it's been tested and proven beneficial
+  loss_gradient_scaling: False
+
+
+  # loss function for the model
+  loss:
+    # loss class to initialise, can be anything subclassing torch.nn.Module
+    _target_: anemoi.training.losses.mse.WeightedMSELoss
+    # what to include in the loss class initialisation
+    include_node_weights: True
```
JesperDramsch, line 50: This could just be `use_node_weights`, then we initialise the loss with it and don't have to do config-shenanigans.

src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
         ----------
         pred : torch.Tensor
-            Prediction tensor, shape (bs, lat*lon, n_outputs)
+            Prediction tensor, shape (bs, (optional_ensemble), lat*lon, n_outputs)
         target : torch.Tensor
-            Target tensor, shape (bs, lat*lon, n_outputs)
+            Target tensor, shape (bs, (optional_ensemble), lat*lon, n_outputs)
         squash : bool, optional
             Average last dimension, by default True
+        feature_indices:
```
JesperDramsch, line 68: Is this for observation code?

src/anemoi/training/losses/mse.py
```diff
@@ -23,69 +24,87 @@ class WeightedMSELoss(nn.Module):
...
-        out *= self.weights[..., None].expand_as(out)
+        # Weight by area, due to weighting construction is analagous to a mean
+        out *= self.node_weights[..., None].expand_as(out)
         # keep last dimension (variables) when summing weights
-        out /= self.sum_function(self.weights[..., None].expand_as(out), axis=(0, 1, 2))
+        out /= self.sum_function(self.node_weights[..., None].expand_as(out))
         return self.sum_function(out, axis=(0, 1, 2))
+
+    @cached_property
+    def name(self) -> str:
```
JesperDramsch, line 109: What is this for?

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,10 +33,34 @@ swa:
 # use ZeroRedundancyOptimizer ; saves memory for larger models
 zero_optimizer: False
 
-# dynamic rescaling of the loss gradient
-# see https://arxiv.org/pdf/2306.06079.pdf, section 4.3.2
-# don't enable this by default until it's been tested and proven beneficial
-loss_gradient_scaling: False
+# loss functions
+loss_functions:
```
JesperDramsch, line 37: This needs a better name as there's only one loss in here.

JesperDramsch (changes requested):
Thanks for tackling these.

I have some doubts about some design decisions, as I have also mentioned in a conversation. 

Could you please elaborate why the loss functions now have a dynamic feature index?

I think the metrics could be a nice list, so we can calculate multiple metrics as well.

src/anemoi/training/losses/mse.py
HCookie: Reverted in next commit back to in-place

src/anemoi/training/losses/mse.py
HCookie: axis is referencing all dimensions of the tensor, there is no need

src/anemoi/training/train/forecaster.py
HCookie: What do you propose for dynamic setting of the initialisation config with calculated arguments?

src/anemoi/training/config/training/default.yaml
HCookie: I can understand the appeal of hard coding things that make sense, but if a user brings a loss function which exposes other configuration options, or deals with nan's another way, we need to support that configuration

HCookie: Setting sensible defaults for our losses will resolve most of the 'bugs' that might be caused by the configurablilty

src/anemoi/training/losses/mse.py
HCookie: Unclear, copied from ensemble codebase

src/anemoi/training/config/training/default.yaml
HCookie: Do you have a suggestion?
As this configures the losses, and metrics it made sense to me

src/anemoi/training/config/training/default.yaml
HCookie: Agreed, will address in next commit

src/anemoi/training/losses/mse.py
HCookie: Addressed in 91e3ae39bd447ed7ef89ad9fdd8af47720ada8af

src/anemoi/training/losses/mse.py
HCookie: Copied from ensemble code base, will identify why

mc4117: It's true that this is the ensemble code but I think we should have renamed this differently.

src/anemoi/training/losses/mse.py
```diff
@@ -10,82 +10,72 @@
...
             Allow nans in the loss and apply methods ignoring nans for measuring the loss, by default False
 
         """
-        super().__init__()
-
-        self.avg_function = torch.nanmean if ignore_nans else torch.mean
-        self.sum_function = torch.nansum if ignore_nans else torch.sum
-
-        self.register_buffer("weights", node_weights, persistent=True)
-        if data_variances is not None:
```
mc4117, line 47: I was just wondering why this has been removed? Has it been added somewhere else?

HCookie: It has been abstracted into the parent `WeightedLoss`

src/anemoi/training/train/forecaster.py
JesperDramsch: Similar pattern to what we use for the Encoder / Decoder / Processor, use config items as config items and code inputs as code inputs.

src/anemoi/training/config/training/default.yaml
JesperDramsch: On another note here, I did consider the "ignore_nans" to be a temporary fix. We could just write two different classes here. One that handles Nans and one that doesn't, then it's handled by the initialisation instead of bloating the function signature.

As a user you can then explicitly choose between the NanWeightedRMSE or the WeightedRMSE, which would now be possible with this implementation.

Rilwan-Adewoyin: > On another note here, I did consider the "ignore_nans" to be a temporary fix. We could just write two different classes here. One that handles Nans and one that doesn't, then it's handled by the initialisation instead of bloating the function signature.
> 
> As a user you can then explicitly choose between the NanWeightedRMSE or the WeightedRMSE, which would now be possible with this implementation.

Yep on the ignore_nans being temporary, I think the initial motivation for having nan_mean and nan_sum in the loss function is also being solved for by PR #72  (pass masks to the loss containing info through where we expect to see nans). 

Unless there was another usecase for the nan_mean/nan_sum, I think they won't be needed after/if PR #72 passes?

src/anemoi/training/train/forecaster.py
```diff
@@ -127,6 +136,53 @@ def __init__(
...
+            )
+
+        loss_config = OmegaConf.to_container(config, resolve=True)
+
+        # Create loss_config including elements from kwargs if they
+        # are specified in the config with include_{key}: True
+        loss_init_config = {}
+        for key in loss_config:  # Go through all keys given in the config
+            # If key does not start with include_, add it to the loss_init_config
+            if not key.startswith("include_"):
```
JesperDramsch, line 173: What is the reason to do this indirectly and not just use the argument directly in the init?

src/anemoi/training/train/forecaster.py
```diff
@@ -263,16 +319,59 @@ def calculate_val_metrics(
         y: torch.Tensor,
         rollout_step: int,
         enable_plot: bool = False,
-    ) -> tuple[dict, list]:
+    ) -> tuple[dict, list[torch.Tensor]]:
+        """Calculate metrics on the validation output.
+
+        Args:
```
JesperDramsch, line 325: Please make sure to set your docstring style to numpy

src/anemoi/training/train/forecaster.py
```diff
@@ -281,7 +380,7 @@ def calculate_val_metrics(
     def training_step(self, batch: torch.Tensor, batch_idx: int) -> torch.Tensor:
         train_loss, _, _ = self._step(batch, batch_idx)
         self.log(
-            "train_wmse",
+            f"train_{getattr(self.loss, 'name', self.loss.__class__.__name__)}",
```
JesperDramsch, line 383: This would not be necessary, if we set the `__repr__` of the class correctly.

src/anemoi/training/train/forecaster.py
```diff
@@ -263,16 +319,59 @@ def calculate_val_metrics(
...
-                y_pred_postprocessed[..., indices],
-                y_postprocessed[..., indices],
-            )
+
+        for metric in self.metrics:
+            metric_name = getattr(metric, "name", metric.__class__.__name__)
+
+            for mkey, indices in self.metric_ranges_validation.items():
+                # for individual variables do not feature scale
+                if len(indices) == 1:
```
JesperDramsch, line 350: If this is only about feature scaling, which it seems to be, this could simply be a boolean `bladibla` and one call `metric(...., feature_scale=bladibla)` instead.

Then it's also clear that `bladibla = mkey == "all"` and nothing more.

src/anemoi/training/train/forecaster.py
```diff
@@ -88,10 +88,19 @@ def __init__(
...
-        self.loss = WeightedMSELoss(node_weights=self.loss_weights, data_variances=loss_scaling)
-        self.metrics = WeightedMSELoss(node_weights=self.loss_weights, ignore_nans=True)
+        loss_kwargs = {"node_weights": self.loss_weights, "feature_weights": loss_scaling}
 
-        if config.training.loss_gradient_scaling:
+        self.loss = self.get_loss_function(config.training.loss_functions.loss, **loss_kwargs)
+        assert isinstance(self.loss, torch.nn.Module) and not isinstance(
+            self.loss,
+            torch.nn.ModuleList,
+        ), "Loss function must be a torch.nn.Module"
```
JesperDramsch, line 97: This assert message needs to be completed.

JesperDramsch (dismissed):
I left a partial review on some of the core problems I see with the current implementation. Thank you for already addressing a lot of my previous comments.

I think there are a few main issues from my side regarding the initialization of the losses. It's likely we could just solve this in a quick call rather than asynchronously.

src/anemoi/training/train/forecaster.py
HCookie: Are you able to review the updated code? It may address your concerns, but if not I am happy to reevaluate.

src/anemoi/training/config/training/default.yaml
HCookie: > I think they won't be needed after/if PR https://github.com/ecmwf/anemoi-training/pull/72 passes?

Okay, thanks @Rilwan-Adewoyin. Could it be PR #72 role to remove the `nan_mean`?

src/anemoi/training/losses/mse.py
HCookie: @mc4117 Do you have a proposal for a better name?

src/anemoi/training/losses/weightedloss.py
```diff
@@ -0,0 +1,159 @@
...
+from abc import abstractmethod
+from functools import cached_property
+
+import torch
+from torch import nn
+
+LOGGER = logging.getLogger(__name__)
+
+
+class WeightedLoss(nn.Module, ABC):
```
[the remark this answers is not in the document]
HCookie: Sure

src/anemoi/training/losses/weightedloss.py
```diff
@@ -0,0 +1,159 @@
...
+        torch.Tensor
+            Weighted loss
+        """
+        out = pred - target
+
+        out = self.scale_by_feature_weights(out, feature_indices, feature_scale)
+        return self.scale_by_node_weights(out, squash)
+
+    @cached_property
+    def name(self) -> str:
```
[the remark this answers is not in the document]
HCookie: I would propose the name of a loss function could be different to its `__str__` and most definitely its `__repr__`.

src/anemoi/training/train/forecaster.py
HCookie: Copied from ensemble codebase, will address

src/anemoi/training/train/forecaster.py
mc4117: hi I think @Rilwan-Adewoyin  would be able to explain this better as it's something that's used in patched energy

src/anemoi/training/train/forecaster.py
```diff
@@ -127,6 +136,53 @@ def __init__(
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         return self.model(x, self.model_comm_group)
 
+    @staticmethod
+    def get_loss_function(
+        config: DictConfig,
+        **kwargs,
+    ) -> Union[torch.nn.Module, torch.nn.ModuleList]:  # noqa: FA100 # stop ruff
```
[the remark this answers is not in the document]
HCookie: Will address in another PR

src/anemoi/training/train/forecaster.py
Rilwan-Adewoyin: tl;dr - for simplicity, the `if len(indices)==1` and `len(indices)>1` logic can be removed, the `elif mkey == "all"` logic can be removed. I would recommend set feature_scale=False and not pass in  feature_indices arg. 

Previously: 
- self.metric_ranges_validation holds mkeys, indices relating to weather variables grouped over pressure levels, and surface variables as a final group.
- in the prev code, when we calculate the loss over the groups (old line 272:275) we use post_processed values 
- I can't remember what the previous behaviour was regarding use of post_processed or non post processed
- When observing validation plots in mlflow, having feature_scaling True or False lead to changes in relative performance of model when looking at performance of these groups (indices) e.g. ModelA>ModelB w/ feature scaling and vice versa otherwise. So this was more useful when I was investigating the perf of different objective funcs. But I don't think it'd be a common usecase at this point so happy for it to be simplified as Jesper proposed. Plus this branch only has the three loss functions, none of which are multivariate so less usecase for configurability here.

HCookie: Thanks for the clarification, the changes have been encorporated

src/anemoi/training/config/training/default.yaml
mc4117: how about removing this higher level, so we have training.loss and training.metrics?

src/anemoi/training/losses/weightedloss.py
```diff
@@ -0,0 +1,159 @@
...
+        error /= self.sum_function(self.node_weights[..., None].expand_as(error), dim=(0, 1, 2))
+        return self.sum_function(error, dim=(0, 1, 2))
+
+    @abstractmethod
+    def forward(
+        self,
+        pred: torch.Tensor,
+        target: torch.Tensor,
+        squash: bool = True,
+        feature_indices: torch.Tensor | None = None,
```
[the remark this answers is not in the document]
mc4117: I don't think this is coming from the observations?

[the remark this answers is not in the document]
floriankrb: I am not sure what functionality we are talking here. 
Please don't mind observations, we can create a WeightedLossForObservations if needed.

mc4117:
This looks good to me. I am happy for this to be approved once the tests have passed

src/anemoi/training/losses/weightedloss.py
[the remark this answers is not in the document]
HCookie: I've checked with the team which included this, it is a useful feature, and is required to be in the call.

src/anemoi/training/losses/weightedloss.py
```diff
@@ -0,0 +1,155 @@
...
+        self,
+        x: torch.Tensor,
+        feature_indices: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Scale a tensor by the loss_scaling.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Tensor to be scaled, shape (bs, (optional_ensemble), lat*lon, n_outputs)
```
mc4117, line 70: We shouldn't have an optional ensemble dimension here. It says optional but actually the squash doesn't work if you don't have an ensemble dimension

HCookie: I believe it is always there, but just sometimes 1

mc4117: OK I think this is maybe not ideal but happy to address this in a subsequent PR

mc4117:
Looks good to me.

HCookie:
Will need to consider a generic way to provide arbitary scalings per dimension.
i.e
```python
variable_scaling, node_scaling, batch_scaling, ensemble_scaling, variable_by_node_scaling, ...

Would be useful to replace loss_scaling

```

mc4117:
We now have a downstream application which depends on this branch so it would be nice if it could be merged soonish

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,11 +33,34 @@ swa:
...
 # dynamic rescaling of the loss gradient
 # see https://arxiv.org/pdf/2306.06079.pdf, section 4.3.2
 # don't enable this by default until it's been tested and proven beneficial
+
+# loss function for the model
+training_loss:
+  # loss class to initialise, can be anything subclassing torch.nn.Module
+  _target_: anemoi.training.losses.mse.WeightedMSELoss
+  # what to include in the loss class initialisation
+  include_node_weights: True
```
mchantry, line 47: could these options be briefly described with a comment please?

src/anemoi/training/losses/combined.py
```diff
@@ -0,0 +1,137 @@
+# (C) Copyright 2024 ECMWF.
```
mchantry, line 1: @HCookie Can you update the headers to be consistent with anemoi-training? (Sorry I realise this has happened while this has been under review)

src/anemoi/training/config/training/default.yaml
```diff
@@ -33,11 +33,34 @@ swa:
...
+  include_variable_scaling: True
+  # other kwargs
+  ignore_nans: False
+
 loss_gradient_scaling: False
 
+# Validation metrics calculation
+validation_metrics:
```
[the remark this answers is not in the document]
HCookie: Addressed in 76947331d5306ab6c9fcd7472e4bb068bd525e62

src/anemoi/training/config/training/default.yaml
HCookie: Addressed in 76947331d5306ab6c9fcd7472e4bb068bd525e62

src/anemoi/training/losses/logcosh.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns
+        -------
+        torch.Tensor
+            Weighted LogCosh loss
+
+        """
+        # Keep logcosh numerically stable
+        x = pred - target
+        s = torch.sign(x) * x
```
mishooax, line 81: is this equivalent to `torch.abs(x)`?

src/anemoi/training/losses/logcosh.py
```diff
@@ -0,0 +1,87 @@
...
+        p = torch.exp(-2 * s)
```
mishooax, line 82: is this op (and its gradient) numerically stable in reduced precision? (no underflow)

src/anemoi/training/losses/logcosh.py
HCookie: Yes, copied from here
https://stackoverflow.com/questions/57785222/avoiding-overflow-in-logcoshx

src/anemoi/training/losses/logcosh.py
HCookie: I'm not sure, nor am I sure how to test it

src/anemoi/training/losses/logcosh.py
```diff
@@ -0,0 +1,87 @@
...
+        feature_scale:
+            If True, scale the loss by the feature_weights
+
+        Returns
+        -------
+        torch.Tensor
+            Weighted LogCosh loss
+
+        """
+        # Keep logcosh numerically stable
```
mishooax, line 79: the derivative of logcosh(x) is tanh(x) - it may be worth exploiting this with a custom autograd function? (to reduce the time spent in backward). e.g. this how it is done in tensorflow
https://github.com/tensorflow/probability/blob/v0.23.0/tensorflow_probability/python/math/generic.py#L766

also, a more general question: are there any advantages to using the logcosh loss over weighted MAE (or MSE)?

src/anemoi/training/losses/logcosh.py
mishooax: okay :) then why not use torch.abs directly (saves us an elementwise op in fwd, and another one in backward)

src/anemoi/training/losses/logcosh.py
mishooax: Did you notice any weird behavior in your tests? I was thinking that `exp(-2x)` may underflow in float16 for large x. FWIW, Tensorflow implements it as `abs(x) + softplus(-2x) - log(2)`
https://github.com/tensorflow/probability/blob/v0.23.0/tensorflow_probability/python/math/generic.py#L754

src/anemoi/training/losses/logcosh.py
HCookie: Thats a good point, a custom autograd could be useful.
Testing has shown that there are some differences, and shows some improvement.

src/anemoi/training/losses/logcosh.py
HCookie: Hadn't noticed anything, but I may not have been running in reduced precision.

HCookie: Would you (@mishooax) suggest copying tensorflow's implementation?

src/anemoi/training/losses/logcosh.py
HCookie: Added in e5fcd926c003cf25ed15c2456bdf70ff43ca90e6

src/anemoi/training/losses/logcosh.py
HCookie: Added in e5fcd926c003cf25ed15c2456bdf70ff43ca90e6

mchantry (approved):
Really nice rework, and great documentation.

PR #76, opened 2024-10-09
Feature/improve dataloader memory

japols:
## Describe your changes
This PR adds a configurable read_group_size (config.dataloader.read_group_size) that defines reader groups, subgroups of the model communication groups, who share the dataloading workload. Increasing the read_group_size heavily reduces CPU memory usage and increases throughput of the dataloader as each task only reads a shard of the train/test/val batch. 

The following experiment on o1280 shows CPU memory usage decrease as we increase the read_group_size (1, 4, 16) with model_sharding across 16 GPUs:

Screenshot 2024-10-30 at 15 15 36

[MLFlow](https://mlflow.ecmwf.int/#/metric?runs=[%2283144751f39a4e09ba73c69f2afcfb98%22,%227127f5eaf6944c26ac13ce4d42e69c8f%22,%22af741eafc6a746dea0f6a754084f533c%22]&metric=%22system%2Fsystem_memory_total_megabytes%22&experiments=[%2234%22]&plot_metric_keys=%5B%22system%2Fsystem_memory_total_megabytes%22%2C%22system%2Fsystem_memory_total_minus_available_megabytes%22%5D&plot_layout={%22autosize%22:true,%22xaxis%22:{},%22yaxis%22:{}}&x_axis=relative&y_axis_scale=linear&line_smoothness=97&show_point=true&deselected_curves=[%227127f5eaf6944c26ac13ce4d42e69c8f-system/system_memory_total_megabytes%22,%22af741eafc6a746dea0f6a754084f533c-system/system_memory_total_megabytes%22]&last_linear_y_axis_range=[])

we can also see a 2x speedup in terms of runtime due to an increased throughput in the dataloader.  

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation and docstrings to reflect the changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have ensured that the code is still pip-installable after the changes and runs
- [x] I have not introduced new dependencies in the inference partion of the model
- [x] I have ran this on single GPU
- [x] I have ran this on multi-GPU or multi-node
- [ ] I have ran this to work on LUMI (or made sure the changes work independently.)
- [x] I have ran the Benchmark Profiler against the old version of the code

## Tag possible reviewers
@ssmmnn11 @mishooax @theissenhelen @JesperDramsch @sahahner @mchantry 

----
📚 Documentation preview 📚: https://anemoi-training--76.org.readthedocs.build/en/76/


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=76) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=76) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=76) it.</sub>

src/anemoi/training/train/forecaster.py
```diff
@@ -124,6 +125,23 @@ def __init__(
             config.hardware.num_gpus_per_node * config.hardware.num_nodes / config.hardware.num_gpus_per_model,
         )
```
ssmmnn11, line 127: Do we actually need any of these here? same for the model_comm_group etc. ... above. I think these are properly initialised by the strategy which uses model.set_model_comm_group and set_reader_group. So might be enough to initialise these to sensible default values when the model is not sharded.

src/anemoi/training/train/forecaster.py
```diff
@@ -219,9 +241,26 @@ def _step(
...
         del batch_idx
+
+        # preprocess batch and broadcast from reader_group rank 0 to reader_group
+        if self.reader_group_rank == 0:
+            # for validation not normalized in-place because remappers cannot be applied in-place
+            batch = self.model.pre_processors(batch, in_place=not validation_mode)
+        else:
+            # init batch tensor with correct shape on non-root ranks
+            shape = (batch.shape[0],) + tuple(batch[0].tolist())
+            batch = torch.zeros(shape, device=self.device)
```
ssmmnn11, line 252: why not torch.empty? Does probably not matter ...

src/anemoi/training/data/datamodule.py
```diff
@@ -74,11 +74,19 @@ def __init__(self, config: DictConfig) -> None:
             * self.config.hardware.num_nodes
             // self.config.hardware.num_gpus_per_model
         )  # number of model communication groups
+
```
ssmmnn11, line 77: It would be nice to get model comm groups and model reader groups from the strategy / use the rountines in the strategy to compute them instead of having code here to re-compute the groups. This should be possible because we initialise the strategy before loading the datamodule.

src/anemoi/training/train/train.py
```diff
@@ -308,6 +308,7 @@ def strategy(self) -> DDPGroupStrategy:
         """Training strategy."""
         return DDPGroupStrategy(
             self.config.hardware.num_gpus_per_model,
+            self.config.dataloader.read_frequency,
```
ssmmnn11, line 311: could we think of a way to make use of the routines / groups computed by the strategy in self.datamodule?

ssmmnn11 (changes requested):
Great work! Would be nice to clean up the group creation a bit more and consolidate everything to the strategy :-).

src/anemoi/training/train/forecaster.py
japols: I changed it to pass everything from the DDPGroupStrategy via model.set_model_comm_group() and set_reader_group().

src/anemoi/training/train/forecaster.py
japols: You're right, no reason to use zeros instead of empty, changed it.

src/anemoi/training/data/datamodule.py
japols: The datamodule only needed these to pass them down to the dataloader.dataset, I removed them entirely from the datamodule now.

src/anemoi/training/train/train.py
japols: I managed to pass them to the dataloader directly via DDPGroupStrategy.process_dataloader() which is called by pytorch_lightning in trainer.fit(model, datamodule).

gabrieloks:
Very nice feature, Jan. I tested this on a rollout run on n320. These runs are painful because we need to reduce the number of workers to avoid out of memory issues and training speed is reduced drastically.

But I did a test with your branch and the develop branch and the results are quite good!

Here is a comparison in terms of memory usage for num_workers = 6:

image

The very good thing is that the job on the develop branch actually crashes at the end of rollout=2 while with your new feature, the rollout fine-tuning keeps going. This will considerably speed up rollout fine-tuning.

japols:
> Very nice feature, Jan. I tested this on a rollout run on n320. These runs are painful because we need to reduce the number of workers to avoid out of memory issues and training speed is reduced drastically.
> 
> But I did a test with your branch and the develop branch and the results are quite good!
> 
> Here is a comparison in terms of memory usage for num_workers = 6:
> 
> image
> 
> The very good thing is that the job on the develop branch actually crashes at the end of rollout=2 while with your new feature, the rollout fine-tuning keeps going. This will considerably speed up rollout fine-tuning.
> 
> @mchantry

Hey Gabriel, happy to hear it helps. Which read_frequency did you use? (4 would probably be optimal)

src/anemoi/training/train/train.py
```diff
@@ -106,7 +105,7 @@ def initial_seed(self) -> int:
         (torch.rand(1), np_rng.random())
         LOGGER.debug(
             "Initial seed: Rank %d, initial seed %d, running with random seed: %d",
-            int(os.environ.get("SLURM_PROCID", "0")),
+            self.strategy.global_rank,
```
mishooax, line 108: is `self.strategy` guaranteed to be correctly initialized at this point?

src/anemoi/training/data/dataset.py
```diff
@@ -93,6 +87,7 @@ def __init__(
         assert self.multi_step > 0, "Multistep value must be greater than zero."
         self.ensemble_dim: int = 2
         self.ensemble_size = self.data.shape[self.ensemble_dim]
+        self.grid_size = self.data.shape[-1]
```
mishooax, line 90: may be useful (easier to understand) if we defined a `self.grid_dim = -1` and used that instead? (like we do for the ensemble dim just above)

mishooax (commented):
Very nice work, Jan! 👍

src/anemoi/training/data/dataset.py
```diff
@@ -128,6 +123,58 @@ def valid_date_indices(self) -> np.ndarray:
...
+        self.model_comm_group_rank = model_comm_group_rank
+        self.model_comm_num_groups = model_comm_num_groups
+        self.reader_group_rank = reader_group_rank
+        self.reader_group_size = reader_group_size
+
+        if self.reader_group_size > 1:
+            # get the grid shard size and start/end indices
+            grid_shard_size = self.grid_size // self.reader_group_size
+            self.grid_start = self.reader_group_rank * grid_shard_size
+            if self.reader_group_rank == self.reader_group_size - 1:
```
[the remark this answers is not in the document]
japols: Not sure it can be shortened since grid_shard_size is `floor(grid_size / reader_group_size)`, so it would have to be a max for the edge case, right?

src/anemoi/training/train/train.py
japols: good point. I think it is since self.strategy is a `@cached_property` i.e. is initialized whenever first called

src/anemoi/training/data/dataset.py
```diff
@@ -233,7 +280,11 @@ def __iter__(self) -> torch.Tensor:
             start = i - (self.multi_step - 1) * self.timeincrement
             end = i + (self.rollout + 1) * self.timeincrement
 
-            x = self.data[start : end : self.timeincrement]
+            if self.reader_group_size > 1:  # read only a subset of the grid
+                x = self.data[start : end : self.timeincrement, :, :, self.grid_start : self.grid_end]
```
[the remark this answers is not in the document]
japols: it's a bit confusing but yes, that's our current understanding. 

CPU memory usage is reduced since each worker now only stores its respective shard of the batch in pinned memory, rather than each worker replicating the entire batch.

If we look at one of the dataloader worker processes, we see that with sharding (left), we spend proportionally less time copying and more time decompressing compared to without sharding (right):

Screenshot 2024-10-28 at 10 11 51

If we were able to decompress only once in shared memory, we would potentially see further improvements.

src/anemoi/training/data/dataset.py
japols: agreed

src/anemoi/training/data/dataset.py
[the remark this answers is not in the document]
mishooax: yes, that's right. good spot

src/anemoi/training/train/train.py
mishooax: 🤔  yes, `Strategy.__init__` gets called when the property is first accessed, but I believe the internal state of the strategy (possibly including `self.strategy.global_rank`) isn't fully initialized until `strategy.setup()` gets called https://github.com/ecmwf/anemoi-training/blob/develop/src/anemoi/training/distributed/strategy.py#L44. And I reckon this doesn't happen until `trainer.fit(...)`?

https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/trainer/trainer.py#L957

japols: you're right - it defaults to 0 before train() gets called. 

The idea was to get rid of all slurm dependencies but maybe in the case of this debug statement we could keep using `SLURM_PROCID`.

mishooax: I think the 0 default is acceptable in this case. The other ranks should log the correct value, after DDP init.

cathalobrien:
> does it make sense to link Jira issues here? or should we manage them independently?
> 
> https://jira.ecmwf.int/browse/HPCAS-53

Probably better to link to here on the Jira ticket, likely more discussion will take place here

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ Keep it human-readable, your future self will thank you!
 - Add anemoi-transform link to documentation
 - Codeowners file (#56)
 - Changelog merge strategy (#56)
+- Feature: Add reader groups to reduce CPU memory usage [#76](https://github.com/ecmwf/anemoi-training/pull/76)
```
HCookie, line 81: This changelog entry will need to be moved.

HCookie (commented):
This looks really good. 
Are there any blockers to merging this soon?

ssmmnn11:
I don't think there is anything outstanding. I agree to merge it rather sooner than later.

ssmmnn11 (dismissed):

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Keep it human-readable, your future self will thank you!
 - Feat: Anemoi Profiler compatible with mlflow and using Pytorch (Kineto) Profiler for memory report [38](https://github.com/ecmwf/anemoi-training/pull/38/)
 - New limited area config file added, limited_area.yaml. [#134](https://github.com/ecmwf/anemoi-training/pull/134/)
 - New stretched grid config added, stretched_grid.yaml [#133](https://github.com/ecmwf/anemoi-training/pull/133)
+- Add reader groups to reduce CPU memory usage and increase dataloader throughput [#76](https://github.com/ecmwf/anemoi-training/pull/76)
```
mchantry, line 49: Due to changelog bot fun, this is in the wrong space, could you put it in the unrelease bit please. (Sorry for the chore)

japols: done: [05665c7](https://github.com/ecmwf/anemoi-training/pull/76/commits/05665c7f8257bf75092049f4f6a71de4055b2fc7)

src/anemoi/training/config/dataloader/native_grid.yaml
```diff
@@ -8,9 +8,9 @@ pin_memory: True
 #   which is then all-gathered between the group.
 #   This can reduce CPU memory usage as well as increase dataloader throughput.
 #   The number of GPUs per model must be divisible by read_group_size.
-#   Good values are num_gpus_per_model or num_gpus_per_node.
+#   To disable, set to 1.
 # ============
-read_group_size: 1
+read_group_size: ${hardware.num_gpus_per_model}
```
mchantry, line 13: This design choice is good for global, but would break distributed LAMs?

ssmmnn11: I think it should be fine. In our sharding, all model instances expect a full batch of data. Here the batch is gathered before it goes to the model. Do we have a simple test setup for lam?

ssmmnn11: If I am looking at the correct branch, 

https://github.com/ecmwf/anemoi-training/blob/4b564e94a35bd028f9d4f54dab25fbb38bf2c8a5/src/anemoi/training/data/dataset.py#L241C1-L242C47

you working with an index. So instead of self.grid_start / self.grid_end, one could compute indices for each reader, either from the full grid or from the subset.

japols: I've added support for LAM in [f08123a](https://github.com/ecmwf/anemoi-training/pull/76/commits/f08123a4f430afd753b2fa08c058a9cb53d5c618). It's a bit weird: anemoi-datasets doesn't support slicing the grid dimension for LAM datasets so I had to load the full grid into memory and then slice the numpy array. This isn't optimal in terms of CPU memory usage but still an improvement. 

We could use this as a preliminary solution and consider updating anemoi-datasets to support slicing LAM data directly in the future.

japols: I've opened a small PR in anemoi-datasets (https://github.com/ecmwf/anemoi-datasets/pull/145) to support the slicing datasets with cutout. When that gets merged we can drop the case distinction from [f08123a](https://github.com/ecmwf/anemoi-training/pull/76/commits/f08123a4f430afd753b2fa08c058a9cb53d5c618).

PR #78, opened 2024-10-09
New mlflow authentication API

gmertes:
This implements the recent changes to the authentication layer for our mlflow server. 

We are no longer asking the user to enter their username/password in a bash prompt because this can be unsafe. Instead, we redirect the user to a webpage where they have to sign in with their ECMWF account and obtain a seed refresh token. The user then copy pastes this token into the prompt.

The rest of the flow remains the same. This manual copy paste action only needs to be done if the user has not logged in within 30 days, just like with the credentials before.

Also made the target env var an optional constructor argument, so it is more clear that this is where the token is stored during runtime.

### DIFF
--- a/src/anemoi/training/diagnostics/mlflow/auth.py
+++ b/src/anemoi/training/diagnostics/mlflow/auth.py
@@ -98,7 +98,7 @@ class TokenAuth:
         Parameters
         ----------
         force_credentials : bool, optional
-            Force a username/password prompt even if a refreh token is available, by default False.
+            Force a credential login even if a refreh token is available, by default False.
         kwargs : dict
             Additional keyword arguments.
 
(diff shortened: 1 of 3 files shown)


src/anemoi/training/diagnostics/mlflow/auth.py
```diff
@@ -98,7 +98,7 @@ def login(self, force_credentials: bool = False, **kwargs: dict) -> None:
         Parameters
```
anaprietonem, line 98: Minor suggestion, but I'd add a comment to the login docstring mentioning that the validity of the refresh token is 30 days. The command provides that information once executed, but I feel like in this piece of code there is no mention about that, so it could be helpful from a developer perspective to know about it

anaprietonem (dismissed):
I left a minor comment but otherwise this PR LGTM and I am happy to approve since it worked successfully both from ATOS and Leonardo!

src/anemoi/training/diagnostics/mlflow/auth.py
gmertes: done!

anaprietonem (approved):

PR #79, opened 2024-10-10
Add output mask

JPXKQX:
This PR adds an output_mask attribute to the GraphForecaster class. This enables rollout training for LAM models and simplifies the diagnostics plot to include only the limited area.

Diagnostic plor

For global models, the previous behaviour is retained where the output_mask is all 1s.

### DIFF
--- a/src/anemoi/training/diagnostics/callbacks/__init__.py
+++ b/src/anemoi/training/diagnostics/callbacks/__init__.py
@@ -689,15 +689,16 @@ class PlotSample(BasePlotCallback):
             ...,
             pl_module.data_indices.internal_data.output.full,
         ].cpu()
-        data = self.post_processors(input_tensor).numpy()
+        data = self.post_processors(input_tensor)
 
         output_tensor = self.post_processors(
             torch.cat(tuple(x[self.sample_idx : self.sample_idx + 1, ...].cpu() for x in outputs[1])),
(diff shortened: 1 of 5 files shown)


src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -689,6 +689,10 @@ def _plot(
             in_place=False,
         ).numpy()
 
+        if pl_module.output_mask is not None:
```
floriankrb, line 692: Perhaps these three lines could be in a method `self.apply_output_mask(data)` ?
_ They would not be duplicated.
_ This would allow more flexibility.

To avoid the two "if" below, you could even have a separate object `self.output_masker.apply_to(data)`

JPXKQX: Thanks for the comment Florian. I have added an apply_output_mask method to fill the values, but not sure how I could avoid the "if"'s.

JPXKQX:
@floriankrb @b8raoult Is there anything else needed to help move this PR forward?

floriankrb (dismissed):
raising a NotImplemetedError instead of "pass" would be nice. But IHMO, this can be merge as it is.

floriankrb (approved):

PR #82, opened 2024-10-11
Fix interactive multi-GPU training

### LINKED ISSUE #80: Able to finish training with one GPU, but got ERROR with "num_gpus_per_node: 2"
### What happened?

After revising ‘num_gpus_per_node’ from 1 to 2 in the debug.yml and run "anemoi-training train --config-name=debug.yaml", I got the error below.

here's my debug.yml

```
defaults:
- data: zarr
- dataloader: native_grid
- diagnostics: eval_rollout
- hardware: example
- graph: multi_scale
- model: gnn
- training: default
- _self_


### This file is for local experimentation.
##  When you commit your changes, assign the new features and keywords
##  to the correct defaults.
# For example to change from default GPU count:
# hardware:
#   num_gpus_per_node: 1

diagnostics:
  plot:
    enabled: False
hardware:
  num_gpus_per_node: -1
  #num_nodes: 1
  #num_gpus_per_model: 2
  paths:
    data: /home/sprixin/test/anemoi
    output: /home/sprixin/test/anemoi/outputs/
    checkpoints: ${hardware.paths.output}checkpoint/
  files:
    dataset: dataset.zarr
    checkpoint:
      every_n_epochs: anemoi-by_epoch-epoch_{epoch:03d}-step_{step:06d}
      every_n_train_steps: anemoi-by_step-epoch_{epoch:03d}-step_{step:06d}
      every_n_minutes: anemoi-by_time-epoch_{epoch:03d}-step_{step:06d}
    warm_start: null
    # graph: ???
  # num_gpus_per_node: 2
  # num_nodes: 1
  # num_gpus_per_model: 1
model:
  num_channels: 128
dataloader:
  limit_batches:
    training: 100
    validation: 100
training:
  max_epochs: 5
```

configuration: 4090*2;  ubuntu 22.04; 
python env: mainly following anemoi documentation

following the anemoi documentation do the training
`anemoi-training train` to generate  debug.yaml and etc.
successful do the first training with debug.yaml(num_gpus_per_node: 1)
get error when changing num_gpus_per_node from 1 to 2

### Version

develop

ubuntu22

### Relevant log output

```shell
...
usage: anemoi-training [-h] [--version] [--debug] {checkpoint,config,mlflow,train} ...
anemoi-training: error: argument command: invalid choice: 'hydra.run.dir="outputs/2024-10-10/11-02-12"' (choose from 'checkpoint', 'config', 'mlflow', 'train')
[2024-10-10 11:02:24,571][lightning_fabric.utilities.distributed][INFO] - Initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/2
[2024-10-10 11:02:29,494][lightning_fabric.strategies.launchers.subprocess_script][INFO] - [rank: 1] Child process with PID 629502 terminated with code 2. Forcefully terminating all other processes to avoid zombies 🧟
Killed
```

### Accompanying data

_No response_

### Organisation

_No response_

gmertes:
This is a proposed fix for interactive multi-GPU training, which is currently broken.

"interactive" in this context means: you call `anemoi-training train` directly in a shell.

### Background

Hydra expects full control over `sys.argv` , with `argv[0]` being the executable, and `argv[1:]` the hydra options.
This is incompatible with our subcommands, where `argv[0]` is `anemoi-training` and `argv[1]` is something like `train`. Hydra will try to ingest `train` as an option, and crash.

To get around this, we overwrite the argv so that `argv[0]` contains the executable and subcommand: `anemoi-training train`.

This makes hydra happy, and it works fine with 1 GPU.

### What works

For most training runs, we run training via Slurm. For this, we don't call the training command directly but we do `srun anemoi-training train` , typically from an sbatch script (but can also be done directly).

The above works fine, because under the hood Slurm will take the argument of srun and execute it once for each GPU. PTL then takes care of the synchronisation between the processes. This can be seen here:

image

One `srun` call results in 2 `anemoi-training train` processes being spawned, one for each GPU (`num_gpus_per_node=2`).

So far so good! 

### What doesn't work

When we run `anemoi-training train` interactively, that is without Slurm, things break. 

This is because without Slurm, PTL will take over responsibility of spawning the subprocesses for each GPU. And as can be seen [here](https://github.com/Lightning-AI/pytorch-lightning/blob/8ad3e29816a63d8ce5c00ac104b14729a4176f4f/src/lightning/fabric/strategies/launchers/subprocess_script.py#L173), it does this by doing a `subprocess.Popen` on the value in `sys.argv[0]`. 

At first glance this looks like it should work, after all `argv[0]` contains `anemoi-training train`. 
But that binary does not exist. So when it tries to open `anemoi-training train` it crashes:

```
/envs/anemoi/bin/python3.10: can't open file '/envs/anemoi/bin/anemoi-training train': [Errno 2] No such file or directory
```

So there is a conflict with the argv. We hack `argv[0]` to make hydra happy, but at the same time it needs to contain a valid binary so that PTL can spawn the DDP processes.

### A solution

One simple solution is to put a value in `argv[0]` that points to a valid binary.

In this PR, I introduce a new "hidden" entry point `.anemoi-training-train` that points to the hydra main function.

Here you can see it working, the top process for GPU 1 is `anemoi-training train`, while the subprocess for GPU 2 is `.anemoi-training-train`:
image

I put the dot at the beginning so that it does not get suggested by autocomplete, because we don't really want users to use this entry point directly. That is also why I added an environment variable to check if the entry point is being called by the subcommand, and not directly.

I then also use that environment variable to log the original command to mlflow. I did not want to touch the top Anemoi trainer interface, so I opted for an env var instead of passing an argument all the way through to the logger.

This is not the cleanest solution. Having this "hidden" entry point is not ideal, so I'm open to suggestions. But I'm in favour of accepting a less-than-ideal solution if it solves the problem in the short run, and then improve it.

This also means that if we have any other subcommands invoking the trainer, we need to have a corresponding "hidden" entry point (e.g.: the profiler).

closes #80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Keep it human-readable, your future self will thank you!
 - Bugfixes for CI (#56)
 - Fix `mlflow` subcommand on python 3.9 [#62](https://github.com/ecmwf/anemoi-training/pull/62)
 - Show correct subcommand in MLFlow - Addresses [#39](https://github.com/ecmwf/anemoi-training/issues/39) in [#61](https://github.com/ecmwf/anemoi-training/pull/61)
+- Fix interactive multi-GPU training [#82](https://github.com/ecmwf/anemoi-training/pull/82)
 
 ### Changed
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,6 +83,10 @@ urls.Documentation = "https://anemoi-training.readthedocs.io/"
 urls.Homepage = "https://github.com/ecmwf/anemoi-training/"
 urls.Issues = "https://github.com/ecmwf/anemoi-training/issues"
 urls.Repository = "https://github.com/ecmwf/anemoi-training/"
+# command for interactive DDP (not supposed to be used directly)
+# the dot is intentional, so it doesn't trigger autocomplete
+scripts.".anemoi-training-train" = "anemoi.training.commands.train:main"
+
 # Add subcommand in the `commands` directory
 scripts.anemoi-training = "anemoi.training.__main__:main"
 
--- a/src/anemoi/training/commands/train.py
+++ b/src/anemoi/training/commands/train.py
@@ -50,6 +50,7 @@ class Train(Command):
         """Merge the sys.argv with the known subcommands to pass to hydra.
 
         This is done for interactive DDP, which will spawn the rank > 0 processes from sys.argv[0]
+        and for hydra, which ingests sys.argv[1:]
 
         Parameters
         ----------
@@ -64,7 +65,7 @@ class Train(Command):
         argv = Path(sys.argv[0])
 
         # this will turn "/env/bin/anemoi-training train" into "/env/bin/.anemoi-training-train"
-        # the dot at the beginning is intentional to not trigger autocomplete
+        # the dot at the beginning is intentional to not interfere with autocomplete
         modified_sysargv = argv.with_name(f".{argv.name}-{args.command}")
 
         if hasattr(args, "subcommand"):
--- a/src/anemoi/training/commands/train.py
+++ b/src/anemoi/training/commands/train.py
@@ -9,7 +9,9 @@
 from __future__ import annotations
 
 import logging
+import os
 import sys
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from anemoi.training.commands import Command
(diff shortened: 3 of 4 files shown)


ssmmnn11:
I tried it on an interactive node and fixes my problem. Great work!

b8raoult (approved):

b8raoult (approved):

PR #86, opened 2024-10-14
Add `AnemoiMlflowClient` with auth support

### LINKED ISSUE #7: Add an AnemoiMlflowClient with authentication support
### Is your feature request related to a problem? Please describe.

If you want to use the mlflow python API on a server with token authentication, you currently have to manually insert the authentication calls before every mlflow call, like this:

```python 
import mlflow
from anemoi.training.diagnostics.mlflow.auth import TokenAuth

tracking_uri = "https://mlflow.url"

auth = TokenAuth(tracking_uri)
client = mlflow.tracking.MlflowClient(tracking_uri)

# check if we already have a login token on disk
# this can also be done from the command line with `anemoi-training mlflow login`
auth.login()

# get an access token
auth.authenticate()
# do mlflow stuff
client.search_experiments()

# authenticate again to be sure the token didn't expire since last time
auth.authenticate()
# do more mlflow stuff
client.log_metric(...)
```

### Describe the solution you'd like

A custom `AnemoiMlflowClient` that inherits from `mlflow.tracking.MlflowClient` , that injects authentication calls before every mlflow call.

```python
from anemoi.training.somewhere import AnemoiMlflowClient

client = AnemoiMlflowClient(tracking_uri, authentication=True)

# do mlflow stuff with login + authentication built in
client.search_experiments()
client.log_metric(...)
```

gmertes:
closes #7 

Implements `AnemoiMlflowClient` that behaves like a regular `mlflow.MlflowClient`, but has the option to inject authentication calls before every mlflow function call. This is to support logging to our server from external codebases, they can import this logger. Authentication is turned off by default.

Usage:
```python
from anemoi.training.diagnostics.mlflow.client import AnemoiMlflowClient

client = AnemoiMlflowClient("http://localhost:5000", authentication=True)

# do regular mlflow client things
client.search_experiments()
client.log_artifact(...)
```

There is also a `check_health` option that queries the health endpoint of the mlflow server on init. We also use this in the anemoi-training logger.

I default this to True because if it fails, it shows a more meaningful error rather than showing a failed HTML request. It reminds the user that maybe they need to authenticate:

`> ConnectionError: Could not connect to MLflow server at http://localhost:5000. The server may require authentication, did you forget to turn it on?`

But it can be turned off. In this mode, the client is completely identical to the mlflow one.

```python 
client = AnemoiMlflowClient("http://localhost:5000", check_health=False)
# is identical to
client = mlflow.MlflowClient("http://localhost:5000")
```

External codebases that don't want to pull the whole dependency stack but just want access to this logger can do:
```
pip install anemoi-training --no-deps
pip install anemoi-utils mlflow
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Keep it human-readable, your future self will thank you!
 - Feature: Add configurable models [#50](https://github.com/ecmwf/anemoi-training/pulls/50)
 - Feature: Authentication support for mlflow sync - [#51](https://github.com/ecmwf/anemoi-training/pull/51)
 - Feature: Support training for datasets with missing time steps [#48](https://github.com/ecmwf/anemoi-training/pulls/48)
+- Feature: `AnemoiMlflowClient`, an mlflow client with authentication support [#86](https://github.com/ecmwf/anemoi-training/pull/86)
 - Long Rollout Plots
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


HCookie (approved):
lgtm

PR #91, opened 2024-10-15
Add expansion of params to logger

### LINKED ISSUE #69: Mlflow log_params string truncation
### What happened?

When logging parameters in mlflow we are truncating the strings to 250 characters. This might be an issue with very long string but also with lists of strings or nested dicts.

Run anemoi-training train with MlFlow tracking enabled

### Version

0.1.0

ATOS

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

_No response_

HCookie:
Closes #69 

Expands iterables into dictionary of form `{key.i value_i}` for mlflow logging

```python
Examples
--------
        >>> expand_iterables({'a': ['a', 'b', 'c']})
        {'a.0': 'a', 'a.1': 'b', 'a.2': 'c', 'a.all': ['a', 'b', 'c'], 'a.length': 3}
        >>> expand_iterables({'a': {'b': ['a', 'b', 'c']}})
        {'a': {'b.0': 'a', 'b.1': 'b', 'b.2': 'c', 'b.all': ['a', 'b', 'c'], 'b.length': 3}}
        >>> expand_iterables({'a': ['a', 'b', 'c']}, size_threshold=100)
        {'a': ['a', 'b', 'c']}
        >>> expand_iterables({'a': [[0,1,2], 'b', 'c']})
        {'a.0': {0: 0, 1: 1, 2: 2}, 'a.1': 'b', 'a.2': 'c', 'a.all': [[0, 1, 2], 'b', 'c'], 'a.length': 3}
```


HCookie:
@MeraX Take a look and see if this addresses the issue you faced.

MeraX:
@HCookie thanks for taking this up. It looks very promising. You are expanding lists/tuples of dicts and dicts recursively. This is fine. I'm wondering if there could be any lists of lists that could be expanded as well. I think list of lists are possible in yaml, but not used in anemoi.

However, I can't test your code it right now, as I have no setup ready that matches the requirements of this branch. Perhaps we could have a quick VC to look at your MLflow output?

The only think I would have chosen differently is the expansion delimiter. I would prefer `.` over `_` as `.` would expand lists the same way as regular dicts (just with integers keys). Alternatively one could also use `f"{key}[{i}]"` as a real index notation.

src/anemoi/training/diagnostics/mlflow/utils.py
```diff
@@ -33,3 +36,79 @@ def health_check(tracking_uri: str) -> None:
...
+
+    nested_func = functools.partial(expand_iterables, size_threshold=size_threshold, recursive=recursive)
+
+    def expand(val: dict | list) -> dict[str, Any]:
+        if not recursive:
+            return val
+        if isinstance(val, dict):
+            return nested_func(val)
+        if isinstance(val, list):
+            return nested_func(dict(enumerate(val)))
```
MeraX, line 97: What about making `expand_iterables` work with a `key, value`-returning iterator? That would save you the conversion to `dict`. In places where you were passing a `dict` before you would then pass `my_dict.items`.

HCookie: It would save the conversion to `dict`, but may reduce the interpretability of the function, and likely result in more conversions to the iterator style.

MeraX: Well, I was just thinking that you are always calling `params.items`, so you could do that also outside. The `params` argument of  `expand_iterables` would then become a 2-tuple iterator. Or the `params` could be a 2-tuple iterator or a `dict` and the `dict` could be immediately turned into a 2-tuple iterator with `dict.items()`. For me, `dict(enumerate(val))` just looks a bit odd. Perhaps a matter of taste.

HCookie:
This expansion of params has been noted to cause performance issues with the mlflow servers, so may need to be considered further

src/anemoi/training/diagnostics/mlflow/utils.py
HCookie: I would prefer to leave it as it is

mchantry:
> This expansion of params has been noted to cause performance issues with the mlflow servers, so may need to be considered further

Is this because the number of logged parameters increases significantly? I think we saw that MLFlow struggles with this.

HCookie:
@mchantry Yeah, increases from 400 -> 4000. when done naively.
With the recent changes that can be brought down to 1200, and further to 500. 
Chatting with @anaprietonem, anything over 1000 is to be avoided. 
Recent changes with this PR will save all configs anyway but into an artifact.

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -492,17 +510,34 @@ def log_hyperparams(self, params: dict[str, Any] | Namespace) -> None:
...
                 truncation_length = 500
-            params_list = [Param(key=k, value=str(v)[:truncation_length]) for k, v in params.items()]
+
+            self.log_hyperparams_as_artifact(expand_iterables(params, size_threshold=truncation_length, delimiter="."))
+
+            expanded_params = {}
+            params = dict(params)
+
+            for key in expand_keys or []:
+                if key in params:
```
anaprietonem, line 527: Could we call 'self.log_hyperparams_as_artifact' over the expanded params to avoid calling the expand_iterables twice?

anaprietonem (commented):
Thanks Harrison! I have left some comments and would be great to see if you have some examples from mlffow runs using this functionality! But thanks so much for the great work!

anaprietonem:
> @mchantry Yeah, increases from 400 -> 4000. when done naively. With the recent changes that can be brought down to 1200, and further to 500. Chatting with @anaprietonem, anything over 1000 is to be avoided. Recent changes with this PR will save all configs anyway but into an artifact.

So with just logging the 'config' expanded we would get to 500? or would we still be in 1200?

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -483,7 +484,24 @@ def _clean_params(params: dict[str, Any]) -> dict[str, Any]:
         return params
 
     @rank_zero_only
-    def log_hyperparams(self, params: dict[str, Any] | Namespace) -> None:
+    def log_hyperparams_as_artifact(self, params: dict[str, Any] | Namespace) -> None:
```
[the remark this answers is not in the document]
HCookie: Sent one to you in private message

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -492,17 +510,34 @@ def log_hyperparams(self, params: dict[str, Any] | Namespace) -> None:
...
             truncation_length = 250
+
             if Version(mlflow.VERSION) >= Version("1.28.0"):
                 truncation_length = 500
-            params_list = [Param(key=k, value=str(v)[:truncation_length]) for k, v in params.items()]
+
+            self.log_hyperparams_as_artifact(expand_iterables(params, size_threshold=truncation_length, delimiter="."))
+
+            expanded_params = {}
+            params = dict(params)
```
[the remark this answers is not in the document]
HCookie: It is already a dict, this is doing a shallow copy, I can update it to be more clear

src/anemoi/training/diagnostics/mlflow/logger.py
HCookie: It is a question of whether we want to the expand config or not, currently it expands all, rather than just those in expand_keys.

anaprietonem (dismissed):
Approving this PR as Harrison has addressed all the comments and suggestions. This changes should allow to have the parameters better logged without over-stressing the server.

anaprietonem (approved):

MeraX:
:+1:

PR #96, opened 2024-10-18
Add ScaleTensor

HCookie:
## ScaleTensor
- Allows dynamic setup of scalars
- Each scalar can be assigned to operate on a dimension or set of dimensions
- Scalar is then resolved at scale time, allowing relative indices

```python
    Examples
    --------
    >>> tensor = torch.randn(3, 4, 5)
    >>> scalars = ScaleTensor((0, torch.randn(3)), (1, torch.randn(4)))
    >>> scaled_tensor = scalars.scale(tensor)
    >>> scalars.get_scalar(tensor.ndim).shape
    torch.Size([3, 4, 1])
    >>> scalars.add_scalar(-1, torch.randn(5))
    >>> scalars.get_scalar(tensor.ndim).shape
    torch.Size([3, 4, 5])
```


src/anemoi/training/losses/weightedloss.py
```diff
@@ -77,16 +90,17 @@ def scale_by_variable_scaling(
...
             return x
 
-        if feature_indices is None:
-            return x * self.variable_scaling
-        return x * self.variable_scaling[..., feature_indices]
+        scalar = self.scalar.get_scalar(x.ndim)
+
+        if feature_indices is None or "variable_scaling" not in self.scalar:
+            return x * scalar
+        return x * scalar[..., feature_indices]
```
floriankrb, line 100: Shouldn't the piece of code be in the ScaleTensor class ?

HCookie: Wdym? The scalar indexing? I would say it depends on the `feature_indices`, so it belongs in the loss function.

PR #97, opened 2024-10-20
Fix pre-commit regex

b8raoult:


HCookie:
The removal of the regex appears to be raising a lot of hidden ruff errors.
Why is this regex being removed? / Will this PR address those new issues?

docs/conf.py
```diff
@@ -16,7 +25,7 @@
 
 read_the_docs_build = os.environ.get("READTHEDOCS", None) == "True"
 
-sys.path.insert(0, os.path.join(os.path.abspath(".."), "src"))
+sys.path.insert(0, os.path.join(os.path.abspath(".."), "src"))  # noqa: PTH118, PTH100
```
theissenhelen, line 28: Didn't we agree on using pathlib instead?

theissenhelen (changes requested):
If possible, please change to pathlib.

JesperDramsch (approved):

PR #83, opened 2024-10-14
Fix/mlflow sync tag

anaprietonem:
This PR includes:
- Fix in mlflow-sync for when we want to sync a new-experiment that doesn't yet exist on the remote server
- Bring a new tag to distinguish between those runs that are synced from offline to online and those are synced 'online' to 'online' (ie between two different remove servers)

PR #98, opened 2024-10-22
Release 0.2.1

gmertes:
Maintenance release to pull in #83. This allows the migrated mlflow runs to be resumed.


gmertes:
Ana found a bug in #83 , waiting for a fix PR.

PR #99, opened 2024-10-22
Fix/mlflow multi gpu

anaprietonem:
Hot-fix for multi-gpu training with mlflow activated on new mlflow production server. 
This error was introduced due to the refactoring done in https://github.com/ecmwf/anemoi-training/pull/83

### DIFF
--- a/src/anemoi/training/diagnostics/mlflow/logger.py
+++ b/src/anemoi/training/diagnostics/mlflow/logger.py
@@ -311,10 +311,10 @@ class AnemoiMLflowLogger(MLFlowLogger):
         self._parent_run_server2server = None
         tags = None
 
-        if rank_zero_only.rank == 0:
-            enabled = authentication and not offline
-            self.auth = TokenAuth(tracking_uri, enabled=enabled)
+        enabled = authentication and not offline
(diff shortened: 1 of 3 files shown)


src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -309,6 +309,7 @@ def __init__(
 
         self._fork_run_server2server = None
         self._parent_run_server2server = None
+        tags = None
```
mchantry, line 312: Why does tags need to be initialised to None?

anaprietonem: Since now the `_get_mlflow_run_params` is a method o `AnemoiMLflowLogger`. Tags are not passed as an input the `AnemoiMLflowLogger` but rather created with this function in the __init__ method. `_get_mlflow_run_params` is just called in rank0 since it's that rank the one that handles of the logging and writing of artifacts, plots and checkpoints. So for all the other ranks we need a default value for the tags to avoid the code failing. (note `run_id` and `run_name` don't need that cause we pass that from the config defaults when defining the `AnemoiMLflowLogger`

gmertes (approved):

PR #102, opened 2024-10-23
Fix/async callbacks

mc4117:
Adding asynchronous callbacks and scatter plots following the former PR by @anaprietonem in aifs-mono (https://github.com/ecmwf-lab/aifs-mono/pull/230/files#diff-817e70e1a151825208d5b55a0b204d08abe52afb9b675b60c48e85376357cc92)

----
📚 Documentation preview 📚: https://anemoi-training--102.org.readthedocs.build/en/102/


src/anemoi/training/config/debug.yaml
```diff
@@ -18,10 +18,13 @@ defaults:
 
 diagnostics:
   plot:
-    enabled: False
-hardware:
-  files:
-    graph: ???
+    enabled: True
```
HCookie, line 21: This should be removed, `plot.enabled` is no longer used

src/anemoi/training/config/debug.yaml
```diff
@@ -18,10 +18,13 @@ defaults:
...
+  log:
+    mlflow:
+      enabled: True
+      authentication: True
+      tracking_uri: https://mlflow.ecmwf.int
+
 model:
```
HCookie, line 28: This should be left as it was, we should not use ecmwf things by default

src/anemoi/training/config/hardware/files/example.yaml
```diff
@@ -1,5 +1,5 @@
-dataset: ???
-graph: ???
+dataset: aifs-ea-an-oper-0001-mars-${data.resolution}-1979-2022-6h-v6.zarr
+graph: graph_anemoi_${data.resolution}.pt
```
HCookie, line 2: This should be left as it was

src/anemoi/training/config/hardware/paths/example.yaml
```diff
@@ -1,6 +1,6 @@
-data: ???
-grids: ???
-output: ???
+grids: /home/mlx/ai-ml/grids/
+data: /home/mlx/ai-ml/datasets/stable/
+output: ${oc.env:SCRATCH}/aifs/${data.resolution}/
 logs:
```
HCookie, line 4: Should be left as it was

HCookie (changes requested):

HCookie:
Given that `scatter` controls if datashader is used, should it be an optional dependency, and import in the plot.py?

src/anemoi/training/config/debug.yaml
mc4117: oh sorry I will change this

mc4117:
> Given that `scatter` controls if datashader is used, should it be an optional dependency, and import in the plot.py?

I guess this is part of a wider conversation whether we want to have the two different options. Thoughts on this @anaprietonem?

CHANGELOG.md
```diff
@@ -74,6 +74,9 @@ Keep it human-readable, your future self will thank you!
 ### Changed
 
 - Updated configuration examples in documentation and corrected links - [#46](https://github.com/ecmwf/anemoi-training/pull/46)
+- Refactored callbacks. [#60](https://github.com/ecmwf/anemoi-training/pulls/60)
+- Remove credential prompt from mlflow login, replace with seed refresh token via web - [#78](https://github.com/ecmwf/anemoi-training/pull/78)
+- Refactored callbacks. [#60](https://github.com/ecmwf/anemoi-training/pulls/60)
```
HCookie, line 79: This is likely to be an issue, can you remove the incorrect updates

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -48,27 +47,6 @@
...
-        return super().submit(self._function_wrapper, fn, *args, **kwargs)
-
-    def _function_wrapper(self, fn: Any, *args: list, **kwargs: dict) -> Callable:
-        """Wraps `fn` in order to preserve the traceback of any kind of."""
-        try:
-            return fn(*args, **kwargs)
-        except Exception as exc:
-            raise sys.exc_info()[0](traceback.format_exc()) from exc
-
-
```
HCookie, line 71: Why has this wrapper been removed?

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -112,27 +100,42 @@ def _output_figure(
...
-                f"{tag}_epoch{epoch:03d}.png",
+                f"{tag}_epoch{epoch:03d}.jpg",
             )
 
             save_path.parent.mkdir(parents=True, exist_ok=True)
-            fig.savefig(save_path, dpi=100, bbox_inches="tight")
+
+            fig.canvas.draw()
+            image_array = np.array(fig.canvas.renderer.buffer_rgba())
+            plt.imsave(save_path, image_array, dpi=100)
```
HCookie, line 110: Can you explain these updates?
They seem seperate to the datashader updates.

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -151,26 +154,33 @@ def _plot(
...
+    # Async function to run the plot function in the background thread
+    async def submit_plot(self, trainer, *args, **kwargs) -> None:
+        """Async function or coroutine to schedule the plot function."""
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(
+            self._executor,
+            self._plot_with_error_catching,
+            trainer,
+            *args,
+            *kwargs.values(),
```
HCookie, line 166: This implies that kwargs will be provided in the correct order?
Why can't this be expanded normally?

src/anemoi/training/diagnostics/maps.py
```diff
@@ -32,7 +32,7 @@ def __init__(self) -> None:
     def __call__(self, lon: np.ndarray, lat: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
         lon_rad = np.radians(lon)
         lat_rad = np.radians(lat)
-        x = [v - 2 * np.pi if v > np.pi else v for v in lon_rad]
+        x = np.array([v - 2 * np.pi if v > np.pi else v for v in lon_rad], dtype=np.float32)
```
HCookie, line 35: Why is this being converted to a numpy array? 
Is this another migrated change?

src/anemoi/training/diagnostics/plots.py
```diff
@@ -473,124 +485,93 @@ def plot_flat_sample(
...
-    """Lat-lon scatter plot: can work with arbitrary grids.
+    """Plot a single lat-lon map.
+
+    Plotting can be made either using scatter plot or Datashader(bin) plots.
+    By default it uses Datashader since it is faster and more efficient.
 
     Parameters
     ----------
-    fig : Figure
+    fig : _type_
```
HCookie, line 574: Can the type be readded?

HCookie (changes requested):

src/anemoi/training/diagnostics/plots.py
anaprietonem: good catch! updating this now. For consistency I am updating all plt.Figure types to just Figure, and there a few more type so I am fixing those to have the correct typings

src/anemoi/training/diagnostics/maps.py
anaprietonem: Yes this is a fix from the PR already opened in aifs-mono. This is basically to ensure that they function type hint is the correct otherwise x would be a list not an array. We also realised that ensuring both x and y are arrays does have a impact on the matplotlib or in general in the plotting generation.

CHANGELOG.md
anaprietonem: should be good now with entry for this PR

src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: This is coming from https://github.com/ecmwf-lab/aifs-mono/pull/230. The aim of that PR was to bring some optimisations for the code to speed up the plotting. Mainly those are:
- Improved asynchronous callback using asyncio (better performance and can handle error catching)
- Introduce the option to use datashader as an alternative to Scatter plots
- Other optimisations like avoiding calling plt.tight_layout or speeding up the saving of plots. 
I can update the PR description with all this info so it helps understand the motivation behind all the changes

src/anemoi/training/diagnostics/callbacks/plot.py
floriankrb: Or maybe something like this ?

```
def func(s, trainer, args, kwargs):
  return s._plot_with_error_catching(trainer, *args, **kwargs)
  
await loop.run_in_executor(
             self._executor,
             func,
             self,
             trainer,
             args,
             kwargs,
 ```
 
 I am not sure we need to pass self as an argument.

Comment (anaprietonem) [reply src/anemoi/training/diagnostics/callbacks/plot.py:174]: So I had a better look into this. And what's happening is that the async function `run_in_executor` doesn't support keyword arguments, so using *kwargs.values() bypasses the need to pass keywords directly by transforming the keyword arguments into positional ones. In that way the values are then correctly used by `_plot_with_error_catching` as positional arguments, which matches the signature that accepts *args.
I have updated the code to reflect that! happy to discuss further!

File: src/anemoi/training/diagnostics/callbacks/plot.py [hunk 11]
```diff
@@ -152,26 +161,34 @@ def _plot(
...
     @rank_zero_only
     def _async_plot(
         self,
-        trainer: pl.Trainer,
-        *args: list,
-        **kwargs: dict,
+        trainer,
+        *args,
+        **kwargs,
     ) -> None:
```
HCookie, line 183: Missing type hints
```suggestion
        trainer: pl.Trainer,
        *args: Any,
        **kwargs: Any,
    ) -> None:
```

src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -152,26 +161,34 @@ def _plot(
...
         """Plotting function to be implemented by subclasses."""
 
+    # Async function to run the plot function in the background thread
+    async def submit_plot(self, trainer, *args, **kwargs) -> None:
+        """Async function or coroutine to schedule the plot function."""
```
HCookie, line 166:
```suggestion
    async def submit_plot(self, trainer: pl.Trainer, *args: Any, **kwargs: Any) -> None:
        """Async function or coroutine to schedule the plot function."""
```

HCookie:
Except the merge conflict LGTM

CHANGELOG.md
```diff
@@ -105,6 +106,7 @@ Keep it human-readable, your future self will thank you!
 
 - Updated configuration examples in documentation and corrected links - [#46](https://github.com/ecmwf/anemoi-training/pull/46)
 - Remove credential prompt from mlflow login, replace with seed refresh token via web - [#78](https://github.com/ecmwf/anemoi-training/pull/78)
+- Modified training configuration to support max_steps and tied lr iterations to max_steps by default [#67](https://github.com/ecmwf/anemoi-training/pull/67)
```
HCookie, line 109: This entry is also duplicated

src/anemoi/training/config/diagnostics/plot/detailed.yaml
```diff
@@ -1,4 +1,5 @@
 asynchronous: True # Whether to plot asynchronously
+scatter: False # Choose which technique to use for plotting
```
HCookie, line 2: Is this supposed to be `datashader`?

src/anemoi/training/config/diagnostics/plot/simple.yaml
```diff
@@ -1,4 +1,5 @@
 asynchronous: True # Whether to plot asynchronously
+scatter: False # Choose which technique to use for plotting
```
HCookie, line 2: Same as other config comment

HCookie (dismissed):

HCookie (approved):

HCookie (approved):

PR #107, opened 2024-10-28
Lock python version to <3.13

HCookie:
## Issue
`pyshtools` cannot be installed from pypi or conda for `3.13`
NOTE: Can be installed from source

## Resolution
Pin version to `<3.13`
- Temporary fix while pyshtools cannot be installed on 3.13

See https://github.com/SHTOOLS/SHTOOLS/issues/494 for more information


mchantry (approved):
LGTM, thanks

PR #114, opened 2024-10-29
MLFlow Offline Failover

HCookie:
- Wraps experiment to capture connection errors
- Logs to disk if server not found

### DIFF
--- a/src/anemoi/training/diagnostics/mlflow/logger.py
+++ b/src/anemoi/training/diagnostics/mlflow/logger.py
@@ -10,6 +10,7 @@
 
 from __future__ import annotations
 
+import functools
 import io
 import logging
 import os
@@ -23,7 +24,10 @@ from typing import Any
 from typing import Literal
 from weakref import WeakValueDictionary
 
+import mlflow
+from mlflow import MlflowClient
 from packaging.version import Version
+from pytorch_lightning.loggers.mlflow import LOCAL_FILE_URI_PREFIX
 from pytorch_lightning.loggers.mlflow import MLFlowLogger
 from pytorch_lightning.loggers.mlflow import _convert_params
 from pytorch_lightning.loggers.mlflow import _flatten_dict
@@ -36,7 +40,6 @@ from anemoi.training.utils.jsonify import map_config_to_primitives
 if TYPE_CHECKING:
     from argparse import Namespace
 
-    import mlflow
 
 LOGGER = logging.getLogger(__name__)
 
@@ -244,6 +247,8 @@ class LogsMonitor:
 class AnemoiMLflowLogger(MLFlowLogger):
     """A custom MLflow logger that logs terminal output."""
 
+    _failed_to_offline = False
+
     def __init__(
         self,
         experiment_name: str = "lightning_logs",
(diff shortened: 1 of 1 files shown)


src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -426,7 +432,71 @@ def _get_mlflow_run_params(
...
+            return
+
+        self._initialized = False
+        self._mlflow_client = MlflowClient(tracking_uri=self._tracking_uri)
+        self._failed_to_offline = False
+
+        from anemoi.training.utils.mlflow_sync import MlFlowSync
+
+        try:
+            MlFlowSync(
```
anaprietonem, line 491: This would try to sync the offline part of the run before logging again online? We'd probably need to use the `self._run_id` and recall `_get_mlflow_run_params` so it restarts from the synced run it

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -323,6 +328,7 @@ def __init__(
             tracking_uri=tracking_uri,
             on_resume_create_child=on_resume_create_child,
         )
+        self._save_dir = f"{LOCAL_FILE_URI_PREFIX}{save_dir}"
```
anaprietonem, line 331: What is the `LOCAL_FILE_URI_PREFIX` doing here?

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -426,7 +432,71 @@ def _get_mlflow_run_params(
     def experiment(self) -> MLFlowLogger.experiment:
         if rank_zero_only.rank == 0:
             self.auth.authenticate()
-        return super().experiment
+
+        parent_obj = super()
+        logger_obj = self
+        experiment = super().experiment
```
anaprietonem, line 438: I think the 'experiment' method does not return an experiment object rather the mlflow client https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/loggers/mlflow.html#MLFlowLogger so what's the idea behind that `experiment=super().experiment` ? This new experiment method won't be returning the same as the original so that could introduce some problems?

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -426,7 +432,71 @@ def _get_mlflow_run_params(
     def experiment(self) -> MLFlowLogger.experiment:
         if rank_zero_only.rank == 0:
             self.auth.authenticate()
-        return super().experiment
+
+        parent_obj = super()
```
anaprietonem, line 436: Just probably me not being familiar with this, could you explain what does `parent_obj = super()` do here?

src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -426,7 +432,71 @@ def _get_mlflow_run_params(
...
+        logger_obj = self
+        experiment = super().experiment
+
+        if self._failed_to_offline:
```
anaprietonem, line 440: What happens if the use is already running in offline mode? That's needed in HPCs like Leonardo and MN5

anaprietonem (commented):
Thanks for this Harrison. Probably we need to iterative over this but could be nice to port this PR over to keep working on it, once the new repo is ready. 

@gmertes we would also need a way to test this functionality. Do you know if we could use the mlflow-test server for that?

PR #116, opened 2024-10-30
fix: metric ranges in the validation space not the normalized space

sahahner:
The wrong output was taken from the `get_val_metric_range` function in the `GraphForecaster`, so that `self.val_metric_ranges` lived in the normalized space. Applying these to the postprocessed fields leads to errors if the remappers are applied.

This PR fixes this.


HCookie (approved):
Thanks for this quick bug fix.
Discussed with Mat, feel free to squash and merge.

Issue #68, opened 2024-10-01
During training, obtaining a resolution of None.


CSyl:
### What happened?

    I have a zarr file that is the subset of the ERA5 zarr: https://console.cloud.google.com/storage/browser/gcp-public-data-arco-era5/ar/1959-2022-1h-360x181_equiangular_with_poles_conservative.zarr?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)). 
    When running anemoi-training train --config-name=debug.yaml, I encountered a "AttributeError: "NoneType" object has no attribute 'lower' " from the anemoi/training/data/datamodule.py line 101 in _check_resolution. 
    Is there a way to verify what the resolution of the https://console.cloud.google.com/storage/browser/gcp-public-data-arco-era5/ar/1959-2022-1h-360x181_equiangular_with_poles_conservative.zarr is in terms of the "o" resolution shown in the anemoi/training/config/data/zarr.yaml file so, that I can set the resolution for training?

### What are the steps to reproduce the bug?

Data & Graph used: Using a anemoi-formatted ERA5 zarr subset extracted from https://console.cloud.google.com/storage/browser/gcp-public-data-arco-era5/ar/1959-2022-1h-360x181_equiangular_with_poles_conservative.zarr and  preprocessed using the anemoi-datasets module & created a graph against the anemoi-formatted ERA5 zarr subset generated using the anemoi-graphs module. 

1) Configuration file (config.yaml) I am using is:

```
defaults:
- data: zarr
- dataloader: native_grid
- diagnostics: eval_rollout
- hardware: example
- graph: multi_scale
- model: gnntransformer
- training: default
- _self_
```

2)  Executed `anemoi-training train  --config-name=config.yaml`
and obtained:

AttributeError: "NoneType" object has no attribute 'lower' " from the anemoi/training/data/datamodule.py line 101 in _check_resolution. 

### Version

0.1.0

### Platform (OS and architecture)

Linux

mchantry:
Hi CSyl
Sorry for the slow reply.
Could you provide access to a small Anemoi dataset -style zarr so we can understand how the resolution has been described in the dataset.
Or a config for anemoi datasets on how you have built the zarr.
Thanks

CSyl:
Hi @mchantry,
    No worries. Thank you for the response. The following is the configuration file which was used to transform the zarr data to an anemoi dataset:
       
**Steps Taken when the Zarr was converted to an anemoi dataset:**

[...]

PR #120, opened 2024-11-01
Update sanity checks for training data consistency

JesperDramsch:
This PR implements a callback that tries to ensure that the variable in the training are in order.

The following scenarios are checked:

## Model Scenarios
There are two scenarios for the model:

### New Model
A fresh model will get the name_to_index from the training dataset and compare to validation and test on epoch start.

### From Checkpoint
A model that is restored from a checkpoint will retrieve the name_to_index in existing data_indices and compare them to instantiated data_indices.

## Error Scenarios
There are valid, warning, and error scenarios:

### Valid Scenario
The name_to_index is identical

### Warnings
When variables are renamed but in the same locations, a warning is issued but training will not be blocked.

### Errors
Values are renamed in different locations, which will result in an error.

Values are permuted, will result in an error

## Removes the Resolution Check

Currently there is a resolution check, which I was told gets in the way of training with merged datasets et al.

Upon consideration of factoring this into another Callback, I decided to get rid of the check instead, as it would be flaky at best and breaking at the worst.

@HCookie pointed out this was also flagged now in #68


src/anemoi/training/data/datamodule.py
HCookie, line 1: I agree this change is needed due to a recent issue reported. 
However, does it fit within this PR?

Additonally, the removal of the resolution check brings into questions the inclusion of resolution in the config, so there may need to be a further discussion about this.

src/anemoi/training/diagnostics/callbacks/sanity.py
```diff
@@ -0,0 +1,168 @@
...
+        Raises
+        ------
+        ValueError
+            If the variable order in the model and data is verifiably different.
+        """
+        if self._model_name_to_index is None:
+            LOGGER.info("No variable order to compare. Skipping variable order check.")
+            return
+
+        if self._model_name_to_index != data_name_to_index:
```
HCookie, line 113: Instead of having a large nested if, it may be simpler to check for equality and exit early

tests/diagnostics/callbacks/test_variable_order.py
HCookie, line 1: A fantastic collection of tests

src/anemoi/training/diagnostics/callbacks/sanity.py
JesperDramsch: Good point. Fixed

tests/diagnostics/callbacks/test_variable_order.py
JesperDramsch: Thanks!

src/anemoi/training/data/datamodule.py
JesperDramsch: Good point. I updated the description and requested comments on a full removal.

HCookie (approved):

HCookie (approved):

HCookie (dismissed):

CHANGELOG.md
```diff
@@ -95,6 +71,7 @@ Keep it human-readable, your future self will thank you!
 - Renamed frequency keys in callbacks configuration. [#118](https://github.com/ecmwf/anemoi-training/pull/118)
 - Modified training configuration to support max_steps and tied lr iterations to max_steps by default [#67](https://github.com/ecmwf/anemoi-training/pull/67)
 - Merged node & edge trainable feature callbacks into one. [#135](https://github.com/ecmwf/anemoi-training/pull/135)
+- Increase the default MlFlow HTTP max retries [#111](https://github.com/ecmwf/anemoi-training/pull/111)
```
mchantry, line 74: Did you mean this to move?

JesperDramsch: Yes, the Changelog had two "Changed" headings, so since I had to fix it anyways, I fixed this as well.

PR #125, opened 2024-11-08
Fix missing checkpoint callbacks

HCookie:
Only one checkpoint callbacks was being added.
This fixes that bug.


JesperDramsch (approved):
Looks like a nice change. LGTM

PR #133, opened 2024-11-11
Graph config for stretched grid graph

havardhhaugen:
**Description**
Adds a config file stretched_grid.yaml intended to be the default when training stretched grid models. 
- Intended to be used with a cutout dataset, and uses the StretchedTriNodes processor mesh with increased resolution over the limited area.
- Processor resolution 5 over the global and 8 over the limited area.
- 12 connections per processor node in the encoder and 3 conndections per input grid node in the decoder.
- Intended to be used with o96 global resolution and 10km limited area resolution for the input grid

Based on the setup used in https://arxiv.org/abs/2409.02891, but with lower resolution. Can be used with e.g. n320 + 2.5km by changing processor resolution to 7+10.

Sample graph using o96 and meps resampled to 10km is attached.

[stretched_grid.zip](https://github.com/user-attachments/files/17700622/stretched_grid.zip)

### DIFF
--- a/src/anemoi/training/config/graph/stretched_grid.yaml
+++ b/src/anemoi/training/config/graph/stretched_grid.yaml
@@ -56,7 +56,7 @@ edges:
 attributes:
   edges:
     edge_length:
-      _target_: anemoi.graphs.edges.attributes.EdgeLenght
+      _target_: anemoi.graphs.edges.attributes.EdgeLength
       norm: unit-max
     edge_dirs:
       _target_: anemoi.graphs.edges.attributes.EdgeDirection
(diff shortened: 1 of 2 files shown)


src/anemoi/training/config/graph/stretched_grid.yaml
```diff
@@ -0,0 +1,63 @@
...
+- source_name: ${graph.hidden}
+  target_name: ${graph.hidden}
+  edge_builder:
+    _target_: anemoi.graphs.edges.MultiScaleEdges
+    x_hops: 1
+  attributes: ${graph.attributes.edges}
+# Decoder
+- source_name: ${graph.hidden}
+  target_name: ${graph.data}
+  edge_builder:
```
JPXKQX, line 51: It seems that there are some spaces instead of tabs or similar that pre-commit does not like. Please run `pre-commit run --all-files' to fix this automatically.

JPXKQX (changes requested):
There are only 2 minor things to change and I can merge,
- the pre-commit issue
- update the changelog with one entry under **Added**. Something like
```
- New stretched grid config added, stretched_grid.yaml. (#133)
```

JPXKQX (approved):

PR #134, opened 2024-11-11
Graph config for limited area models (LAMs)

JPXKQX:
This PR includes a base config file for creating graphs over an area of interest. It also pins the anemoi-graphs version to be at least 0.4 which introduces the support for stretched grids and limited area graphs.


theissenhelen (approved):
Good work!

PR #64, opened 2024-09-27
Make pin_memory of the Dataloader configurable

MeraX:
Pinned memory caused crashes on our machine, so we need this setting to be configurable.

PR #135, opened 2024-11-11
Feature/update graph callbacks

JPXKQX:
This PR updates the learned features callbacks to work with new version of anemoi-models ([ PR #64](https://github.com/ecmwf/anemoi-models/pull/64)). In addition, it merges both Node & Edge callbacks into a single `GraphTrainableFeaturesCallback` callback.


JPXKQX:
The split of the callbacks was related to PlotHistograms and PlotSpectrum. The `GraphEdgeTrainableCallback` was created independently, but at least in my case it is something that I only run with `GraphNodeTrainableCallback`. Happy to hear other people's views

PR #130, opened 2024-11-08
Feature: MSE metrics inside/outside regional domain for stretched grid models (KNMI)

jswijnands:
_Description of functionality_
Four new metrics have been added for stretched grid models. The first two metrics compute the weighted MSE (i) only inside the regional domain and (ii) only outside the regional domain. The former will allow for comparison of LAM and stretched grid models in the future. Further, two additional metrics have been added that compute the contributions of these components to the overall weighted MSE. This enables monitoring which of the two is the main driver of the training process, as it indicates whether we are mainly optimising the loss inside or outside the regional domain.

_Configuration options_
Both features can be turned on or off in the config options by specifying the following stretched grid metrics (sg_metrics) in the diagnostics config:
- wmse_per_region (True/False)
- wmse_contributions (True/False)

_Outputs_
Plots of these metrics are generated in MLFlow, specified as  ‘wmse_inside_lam_epoch’, ‘wmse_outside_lam_epoch’, ‘wmse_inside_lam_contribution_epoch’, and ‘wmse_outside_lam_contribution_epoch’.

MLFlow plots w-MSE splitting_default_area_weighting

_Further comments_
- Note that this PR is dependent on a PR that provides the stretched_grid.yaml graph config.
- This PR has also been tested together with the PR on custom domain weighting, to check that these features interact as expected. 
- A CERRA config has been added for testing purposes, but does not have to be merged into anemoi.

PR #136, opened 2024-11-11
Feature: Option to set / change area weighting outside of graph-creation

havardhhaugen:
**Desctiption of functionality**
Adds functionality to change the area weighting of nodes in the limited area of a stretched grid graph.
Adds two new classes to scaling.py:
- BaseAreaWeights:
Uses scipy.SphericalVoronoi to calculate area weights based on latlon input grid similarly to how this is currently done in anemoi-graphs. Currently this probably won't have many use-cases since the area weights are already calculated like this in graphs, but it serves as a good base class.
- StretchedGridCutoutAreaWeights:
Re-calculates area-weights using BaseAreaWeights and then scales the area weights in the limited area (cutout) such that their sum is equal to a configurable fraction of the sum of global area weights. 

Adds staticmethod get_node_weights to GraphForecaster which uses the new scaling methods if they are set in the config.

**Configuration options**
Specified in config.training.loss_scaling.spatial, for instance:
training:
  loss_scaling:
    spatial:
      _target_: anemoi.training.data.scaling.StretchedGridCutoutAreaWeights
      target_nodes: ${graph.data}
      cutout_weight_frac_of_global: 0.3 
      
**Further comments**
- Based on the implementation of area-weights in the stretched grid branch of aifs-mono and https://github.com/ecmwf/anemoi-training/tree/feature/limited_area 
-  Functionality tested (together with stretched grid specific metrics from #130) https://mlflow.ecmwf.int/#/metric?runs=[%227c31b5334f524b78a74bbdd99d6f4bf6%22,%22e03701c4613745a884c9aae34e83d472%22]&metric=%22val_wmse_inside_lam_contribution_epoch%22&experiments=[%2230%22]&plot_metric_keys=%5B%22val_wmse_inside_lam_contribution_epoch%22%5D&plot_layout={%22autosize%22:true,%22xaxis%22:{%22range%22:[-3752.513985943475,19833.295466415468],%22autorange%22:false},%22yaxis%22:{%22range%22:[-0.01092616814456952,0.04189337908390672],%22autorange%22:false}}&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]

Thanks to Magnus Ingstad, Jasper Wijnands and Sophie Buurman and Mario Santa Cruz for their contributions to this PR.

### DIFF
--- a/src/anemoi/training/data/scaling.py
+++ b/src/anemoi/training/data/scaling.py
@@ -13,11 +13,10 @@ from abc import ABC
 from abc import abstractmethod
 
 import numpy as np
-
 import torch
-from torch_geometric.data import HeteroData
-from scipy.spatial import SphericalVoronoi
 from anemoi.graphs.generate.transforms import latlon_rad_to_cartesian
+from scipy.spatial import SphericalVoronoi
+from torch_geometric.data import HeteroData
 
 LOGGER = logging.getLogger(__name__)
 
@@ -83,6 +82,7 @@ class NoPressureLevelScaler(BasePressureLevelScaler):
         # no scaling, always return 1.0
         return 1.0
 
+
 class BaseAreaWeights:
     """Method for overwriting the area-weights stored in the graph object."""
 
@@ -96,29 +96,34 @@ class BaseAreaWeights:
         """
         self.target = target_nodes
 
-    def area_weights(self, graph_data) -> torch.Tensor:
+    def area_weights(self, graph_data: HeteroData) -> torch.Tensor:
         return torch.from_numpy(self.global_area_weights(graph_data))
 
     def global_area_weights(self, graph_data: HeteroData) -> np.ndarray:
-        lats, lons = graph_data[self.target].x[:,0], graph_data[self.target].x[:,1]
+        lats, lons = graph_data[self.target].x[:, 0], graph_data[self.target].x[:, 1]
         points = latlon_rad_to_cartesian((np.asarray(lats), np.asarray(lons)))
         sv = SphericalVoronoi(points, radius=1.0, center=[0.0, 0.0, 0.0])
         area_weights = sv.calculate_areas()
 
         return area_weights / np.max(area_weights)
 
+
 class StretchedGridCutoutAreaWeights(BaseAreaWeights):
-    """Rescale area weight of nodes inside the cutout area by setting their sum to a fraction of the sum of global nodes area weight."""
+    """Rescale area weight of nodes inside the cutout area.
+
+    Sum of the area weight of cutout nodes set to a specified fraction of sum of global nodes.
+    """
 
     def __init__(self, target_nodes: str, cutout_weight_frac_of_global: float):
         """Initialize area weights with target nodes and scaling factor for the cutout nodes area weight.
-        
+
         Parameters
         ----------
         target_nodes : str
             Name of the set of nodes to be rescaled (defined when creating the graph).
         cutout_weight_frac_of_global: float
-            Scaling factor for the cutout nodes area weight - sum of cutout nodes area weight set to a fraction of the sum of global nodes area weight.     
+            Scaling factor for the cutout nodes area weight - sum of cutout nodes area weight set to a fraction of
+            the sum of global nodes area weight.
         """
         super().__init__(target_nodes=target_nodes)
         self.fraction = cutout_weight_frac_of_global
--- a/src/anemoi/training/data/scaling.py
+++ b/src/anemoi/training/data/scaling.py
@@ -14,6 +14,11 @@ from abc import abstractmethod
 
 import numpy as np
 
+import torch
+from torch_geometric.data import HeteroData
+from scipy.spatial import SphericalVoronoi
+from anemoi.graphs.generate.transforms import latlon_rad_to_cartesian
+
 LOGGER = logging.getLogger(__name__)
 
 
@@ -77,3 +82,53 @@ class NoPressureLevelScaler(BasePressureLevelScaler):
         del plev  # unused
         # no scaling, always return 1.0
         return 1.0
+
+class BaseAreaWeights:
+    """Method for overwriting the area-weights stored in the graph object."""
+
+    def __init__(self, target_nodes: str):
+        """Initialize area weights with target nodes.
+
+        Parameters
+        ----------
+        target_nodes : str
+            Name of the set of nodes to be rescaled (defined when creating the graph).
+        """
+        self.target = target_nodes
+
+    def area_weights(self, graph_data) -> torch.Tensor:
+        return torch.from_numpy(self.global_area_weights(graph_data))
+
+    def global_area_weights(self, graph_data: HeteroData) -> np.ndarray:
+        lats, lons = graph_data[self.target].x[:,0], graph_data[self.target].x[:,1]
+        points = latlon_rad_to_cartesian((np.asarray(lats), np.asarray(lons)))
+        sv = SphericalVoronoi(points, radius=1.0, center=[0.0, 0.0, 0.0])
+        area_weights = sv.calculate_areas()
+
+        return area_weights / np.max(area_weights)
+
+class StretchedGridCutoutAreaWeights(BaseAreaWeights):
+    """Rescale area weight of nodes inside the cutout area by setting their sum to a fraction of the sum of global nodes area weight."""
+
+    def __init__(self, target_nodes: str, cutout_weight_frac_of_global: float):
+        """Initialize area weights with target nodes and scaling factor for the cutout nodes area weight.
+        
+        Parameters
+        ----------
+        target_nodes : str
+            Name of the set of nodes to be rescaled (defined when creating the graph).
+        cutout_weight_frac_of_global: float
+            Scaling factor for the cutout nodes area weight - sum of cutout nodes area weight set to a fraction of the sum of global nodes area weight.     
+        """
+        super().__init__(target_nodes=target_nodes)
+        self.fraction = cutout_weight_frac_of_global
+
+    def area_weights(self, graph_data: HeteroData) -> torch.Tensor:
+        area_weights = self.global_area_weights(graph_data)
+        mask = graph_data[self.target]["cutout"].squeeze().bool()
+
+        global_sum = np.sum(area_weights[~mask])
+        weight_per_cutout_node = self.fraction * global_sum / sum(mask)
+        area_weights[mask] = weight_per_cutout_node
+
+        return torch.from_numpy(area_weights)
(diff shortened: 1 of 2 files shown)


JPXKQX (commented):
Thanks for the contributions!, I really like making the spatial scaling explicit. I think it may be a good opportunity to refactor this `node_loss_weights` into a new config field under `training`. I leave a few things open for discussion below:
- Merge the model.node_loss_weights and the training.loss_scaling.spatial and taking it outside the `training.loss_scaling` to `training.node_loss_weights` ??.
- I am thinking of something like the following (names to be discussed):
```
node_loss_weights:
      _target_: anemoi.training.losses.scaling.ReweightedGraphNodeAttribute
      target_nodes: ${graph.data}
      node_attribute: area_weight   
      scale_by_attribute: "cutout"
      weight_frac_of_total: 0.3     
```
- Make it explicit for the global model too, so we have something like:
```
node_loss_weights:
      _target_: anemoi.training.losses.scaling.GraphNodeAttribute
      target_nodes: ${graph.data}
      node_attribute: area_weight     
```
- Maybe the code fits better in the `losses/` folder, what do you think?

src/anemoi/training/data/scaling.py
```diff
@@ -77,3 +81,59 @@ def scaler(plev: float) -> np.ndarray:
...
+        ----------
+        target_nodes : str
+            Name of the set of nodes to be rescaled (defined when creating the graph).
+        """
+        self.target = target_nodes
+
+    def area_weights(self, graph_data: HeteroData) -> torch.Tensor:
+        return torch.from_numpy(self.global_area_weights(graph_data))
+
+    def global_area_weights(self, graph_data: HeteroData) -> np.ndarray:
```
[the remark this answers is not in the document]
havardhhaugen: The intention was not to have BaseAreaWeights as a default in forecaster to always re-calculate the areaweights in the graph. The default in forecaster is still to load the area-weights from the graph, which is achieved by not specifying training.loss_scaling.spatial in the config. 
I included this because it seemed like an appropriate base class for the cutout-reweighing (and potentially others to come), and for the very limit use-case when someone has made a graph with (for example) uniform weights and wish to overwrite those.

src/anemoi/training/train/forecaster.py
```diff
@@ -290,6 +290,17 @@ def get_feature_weights(
...
+    def get_node_weights(config: DictConfig, graph_data: HeteroData) -> torch.Tensor:
+        node_weights = graph_data[config.graph.data][config.model.node_loss_weight].squeeze()
+
+        if "spatial" in config.training.loss_scaling:
+            spatial_loss_scaler = instantiate(config.training.loss_scaling.spatial)
+            node_weights = spatial_loss_scaler.area_weights(graph_data)
+            LOGGER.info("Rescaling area weights")
+
+        return node_weights
+
```
HCookie, line 303: As we discussed, I like the idea of a way to scale/modify the node weights. 
So I think that this function can be config controlled, and allow arbitary modifications to be applied.

havardhhaugen:
I updated the implementation based on your feedback. The base class behavior is now to load weights from the graph with a fall-back of calculating area-weights on the fly. Similiarly ReweightedGraphNodeAttribute will also attempt to load the weights from the graph and then re-weights. 

I also changed the re-weighting implementation slightly to be more in line with https://github.com/ecmwf/anemoi-training/blob/feature/limited_area/src/anemoi/training/train/forecaster.py. In aifs-mono we scaled the limited area nodes to be a fraction of the global (unmasked) sum, but since the function is now more generic it makes more sense to use a fraction of the total (masked + unmasked).

I didn't  actually test if the code works as expected yet, will do that when we are happy with the intended behaviour.

src/anemoi/training/config/training/default.yaml
```diff
@@ -112,3 +112,8 @@ pressure_level_scaler:
   _target_: anemoi.training.data.scaling.ReluPressureLevelScaler
   minimum: 0.2
   slope: 0.001
+
+node_loss_weights:
+  _target_: anemoi.traininig.losses.nodeweights.GraphNodeAttribute
```
HCookie, line 117:
```suggestion
  _target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
```

HCookie:
It's looking very good, I like the overall implementation, and think it will become very useful in the future. 
I've left a few initial comments to start with.

havardhhaugen:
Fixed the issues addressed in the comments and made sure the weights are loaded correctly in training for the four cases: load from graph (+fallback) and reweight (+fallback).

src/anemoi/training/losses/nodeweights.py
```diff
@@ -0,0 +1,78 @@
...
+import torch
+from anemoi.graphs.generate.transforms import latlon_rad_to_cartesian
+from scipy.spatial import SphericalVoronoi
+from torch_geometric.data import HeteroData
+
+LOGGER = logging.getLogger(__name__)
+
+
+class GraphNodeAttribute:
+    """Method to load and optionally change the weighting of node attributes in the graph."""
```
HCookie, line 23: Given this is the base class, It would be good if all functions had complete docstrings.

src/anemoi/training/losses/nodeweights.py
```diff
@@ -0,0 +1,78 @@
...
+            LOGGER.info("Loading node attribute %s from the graph", self.node_attribute)
+        except KeyError:
+            attr_weight = torch.from_numpy(self.area_weights(graph_data))
+
+            LOGGER.info(
+                "Node attribute %s not found in graph. Default area weighting will be used",
+                self.node_attribute,
+            )
+
+        return attr_weight
```
HCookie, line 50: Would it be possible to do a `if x in y` rather than a try except?

src/anemoi/training/losses/nodeweights.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, target_nodes: str, node_attribute: str, scaled_attribute: str, weight_frac_of_total: float):
+        super().__init__(target_nodes=target_nodes, node_attribute=node_attribute)
+        self.scaled_attribute = scaled_attribute
+        self.fraction = weight_frac_of_total
+
+    def weights(self, graph_data: HeteroData) -> torch.Tensor:
+        attr_weight = super().weights(graph_data)
+
+        mask = graph_data[self.target][self.scaled_attribute].squeeze().bool()
```
HCookie, line 68: An `if in` may help to raise more helpful errors here.

havardhhaugen:
Made docstrings and changed the try-except. Should I remove ```node_loss_weight: area_weight``` from the model config files as part of this pr?

HCookie:
> Should I remove node_loss_weight: area_weight 
Why? Given your updates to the forecaster, it is now a mandatory config

src/anemoi/training/losses/nodeweights.py
```diff
@@ -0,0 +1,129 @@
...
+
+class ReweightedGraphNodeAttribute(GraphNodeAttribute):
+    """Method to reweight a subset of the target nodes defined by scaled_attribute.
+
+    Subset nodes will be scaled such that their weight sum equals weight_frac_of_total of the sum
+    over all nodes.
+    """
+
+    def __init__(self, target_nodes: str, node_attribute: str, scaled_attribute: str, weight_frac_of_total: float):
+        super().__init__(target_nodes=target_nodes, node_attribute=node_attribute)
```
HCookie, line 111: Please add a docstring to this function

src/anemoi/training/losses/nodeweights.py
HCookie, line 1: If possible, could you add some tests to ensure the correct weightings are used?

havardhhaugen: I have not looked into how you do testing so its not obvious to me how these tests should look. Do you have a set of benchmark graphs (or recipe files to create them) that I should test the functionality on? Or is this something I should create as part of the tests?

HCookie: Take a look at the tests directory in training and in graphs. 
Considering you are not passing the config to the classes, you can likely mock a `HeteroData`, run `weights` and check the returning value. Make sure to test as many edge cases as you can think of. 
If you would like to chat, we can have a meeting

havardhhaugen: I've added some tests that should cover the current use cases of the added functionality.

src/anemoi/training/config/training/default.yaml
```diff
@@ -115,3 +115,8 @@ pressure_level_scaler:
   _target_: anemoi.training.data.scaling.ReluPressureLevelScaler
   minimum: 0.2
   slope: 0.001
+
+node_loss_weights:
+  _target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
```
mchantry, line 120: Would be great to see this, and the stretched-grid option documented in the docs.

mchantry:
Really nice contribution, thanks.

havardhhaugen:
I added documentation in user-guide.training under loss function scaling.

src/anemoi/training/losses/nodeweights.py
```diff
@@ -0,0 +1,143 @@
...
+        ----------
+        target_nodes: str
+            name of nodes, key in HeteroData graph object
+        node_attribute: str
+            name of node weight attribute, key in HeteroData graph object
+        """
+        self.target = target_nodes
+        self.node_attribute = node_attribute
+
+    def area_weights(self, graph_data: HeteroData) -> np.ndarray:
```
JPXKQX, line 52: Ophelia recently solved the scipy.Voronoi issue. I think it would be nice to compute the area weights using the same method as in anemoi-graphs. I think this method can be changed to:
```
from anemoi.graphs.nodes.attributes import AreaWeights

AreaWeights(norm="unit-max", fill_value=0).get_raw_values(graph[self.target])
```

havardhhaugen: good idea! I've changed it and updated the pytests to also use AreaWeights from graph

src/anemoi/training/losses/nodeweights.py
HCookie: Yeah I do

havardhhaugen:
This should be finished now. I can see that it is failing some of the pytests, but I think that is because these are being run with an older version of anemoi-graphs.

JPXKQX:
PR #159 will bump the minimum required version of `anemoi-graphs` to `0.4.1`.

JPXKQX:
@havardhhaugen can you rebase `develop` branch? `anemoi-graphs=0.4.1` should be a requirement now.

HCookie (approved):

PR #137, opened 2024-11-11
fix: Allow updates to scalars & other improvements

HCookie:
# Scalar updates

Allow the ScaleTensor of the loss function to have elements updated.

```python
ScaleTensor(test = (-1, torch.ones((2,))).update(test = torch.ones((2,)) * 3)

WeightedMSELoss().update_scalar('test', 3)
```

Additionally, add a `without` to allow subsetting without the given scalars.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Keep it human-readable, your future self will thank you!
 ## [Unreleased](https://github.com/ecmwf/anemoi-training/compare/0.2.2...HEAD)
 
 ### Fixed
+- Allow updates to scalars
 - Refactored callbacks. [#60](https://github.com/ecmwf/anemoi-training/pulls/60)
     - Updated docs [#115](https://github.com/ecmwf/anemoi-training/pull/115)
     - Fix enabling LearningRateMonitor [#119](https://github.com/ecmwf/anemoi-training/pull/119)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Keep it human-readable, your future self will thank you!
 - Sub-hour datasets [#63](https://github.com/ecmwf/anemoi-training/pull/63)
 - Add synchronisation workflow [#92](https://github.com/ecmwf/anemoi-training/pull/92)
 - Feat: Anemoi Profiler compatible with mlflow and using Pytorch (Kineto) Profiler for memory report [38](https://github.com/ecmwf/anemoi-training/pull/38/)
+- New limited area config file added, limited_area.yaml. [#134](https://github.com/ecmwf/anemoi-training/pull/134/)
 - New stretched grid config added, stretched_grid.yaml [#133](https://github.com/ecmwf/anemoi-training/pull/133)
 
 ### Changed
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Keep it human-readable, your future self will thank you!
 ## [Unreleased](https://github.com/ecmwf/anemoi-training/compare/0.2.2...HEAD)
 
 ### Fixed
+- Allow updates to scalars
```
HCookie, line 14:
```suggestion
- Allow updates to scalars [#137](https://github.com/ecmwf/anemoi-training/pulls/137)
```

sahahner:
What would be a use case for the `without` and `without_by_dim` functionalities that are provided by this PR?

HCookie:
> What would be a use case for the `without` and `without_by_dim` functionalities that are provided by this PR?

I have just added a use case, 
In the verification metrics we can chose to remove scalars in the calculation.
It will also become useful in a later PR for stretched grids

src/anemoi/training/losses/weightedloss.py
```diff
@@ -64,30 +65,45 @@ def __init__(
...
         """Scale a tensor by the variable_scaling.
 
         Parameters
         ----------
         x : torch.Tensor
             Tensor to be scaled, shape (bs, ensemble, lat*lon, n_outputs)
-        feature_indices:
-            feature indices (relative to full model output) of the features passed in pred and target
+        feature_indices: torch.Tensor, optional
+            feature indices (relative to full model output) of the features/variables passed in pred and target
```
sahahner, line 86: Should this be relative to the internal model output? I think we should follow the namespace of the data indices here.

sahahner: I noticed that for the validation metrics, "relative to the model output" is indeed correct. Either way, we should find a way to keep this description general.

HCookie: Agreed, we should keep the names consistent.
This is code transferred from the old losses, so not attached. I'll discuss with @Rilwan-Adewoyin who added this

HCookie: > to keep this description general.

Do you have a suggestion?

sahahner:
>  In the verification metrics we can chose to remove scalars in the calculation. 

Is it also possible to apply this functionality on the training loss?

src/anemoi/training/losses/weightedloss.py
sahahner: `feature indices (relative to model output to which this loss is applied) of the features/variables passed in pred and target`

HCookie:
> Is it also possible to apply this functionality on the training loss?

Techically possible, yes, would we want that behaviour tho?

sahahner:
> > Is it also possible to apply this functionality on the training loss?
> 
> Techically possible, yes, would we want that behaviour tho?

If it were not possible, I would suggest an assert. But if technically possible we can leave it that general.

src/anemoi/training/losses/weightedloss.py
sahahner: I like the new name and docstring. We can update this one as well to `Indices to subset the calculated scalar with, by default None`.

sahahner (approved):
LGTM. Nice functionality that directly will be used in other PRs.

PR #138, opened 2024-11-11
fix: Rename loss_scaling to variable_loss_scaling

HCookie:
- Update configs to tell user where the variable scalars is defined


mc4117:
This looks good but I think you might also need to rename in the documentation?

mc4117 (approved):

PR #141, opened 2024-11-13
feat: Add randomiser option to rollout

HCookie:
# Allow randomised rollout when training.

## Applications
- Prevent a model overfitting to a certain rollout
- Tune for various times

## Usage
Config file
```yaml
training:
  rollout:
    # randomise rollout length during training
    randomise: True
```

----
📚 Documentation preview 📚: https://anemoi-training--141.org.readthedocs.build/en/141/


jakob-schloer:
I did something similar back in aifs-mono which I have never ported to anemoi-training: 
https://github.com/ecmwf-lab/aifs-mono/tree/feature/rollout-strategies

Would be nice to have different rollout strategies.

Issue #126, opened 2024-11-08
PlotSpectrum does not work for LAM


JPXKQX:
### What happened?

PlotSpectrum callback fails when included in LAM experiments.

### What are the steps to reproduce the bug?

Train a limited area model with `plot: detailed` so PlotSpectrum is used.

### Version

develop

### Platform (OS and architecture)

x86_64 GNU/Linux

### Relevant log output

```shell
File ".../anemoi-training/src/anemoi/training/diagnostics/plots.py", line 255, in compute_spectra
    coeffs_field = SHExpandGLQ(field, w=zero_w[1], zero=zero_w[0])
  File ".../conda/envs/anemoi-dev/lib/python3.10/site-packages/pyshtools/backends/shtools.py", line 212, in wrapped_func
    raise SHToolsError(_shtools_status_message(returned_values[0]))
pyshtools.backends.shtools.SHToolsError: Improper dimensions of input array.
```
```


### Accompanying data

_No response_

### Organisation

ECMWF

lattaai4:
Hello, I tried to fix the issue

This is what I did:

Modify the compute_spectra function to use numpy's FFT for LAM grids instead of spherical harmonics. This change allows the function to work with both global and limited area models.

I also created pull request: https://github.com/ecmwf/anemoi-training/pull/127


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

PR #149, opened 2024-11-18
[FIX] Power spectra bug on n320 (LAM?)

gabrieloks:
Related to #126.

To plot the power spectra, we need to create a regular grid (n_pix_lat x n_pix_lon) and interpolate the data on it. The way n_pix_lat and n_pix_lon were previously defined is not robust and might lead to errors.

I am more or less confident that this change should fix the issue on n320 but not sure about LAM.


japols (dismissed):
looks good and works for me (on n320)

HCookie:
@gabrieloks Please resolve the pre-commit errors and a changelog.
Then we can look at merging

Issue #150, opened 2024-11-20
N320 callback Number of pixels used by datashader


anaprietonem:
### What happened?

Current value of 'n_pixels' variable produce too sparse plots when working with n320.

### What are the steps to reproduce the bug?

Run debug run with n320 with asynchronous and datashader set to True 

<img width="1038" alt="Screenshot 2024-11-19 at 15 18 45" src="https://github.com/user-attachments/assets/16096321-91c0-4484-b90b-35af1b561315">

### Version

0.3

### Platform (OS and architecture)

ATOS

PR #152, opened 2024-11-20
150 n320 callback number of pixels used by datashader

anaprietonem:
PR to fix https://github.com/ecmwf/anemoi-training/issues/150
Propose change - make the 'n_pixels' more flexible so that  when using finer resolutions (like n32),  500 as `n_pixels` should be enough, but for hidden grid plots and coarser ones then we need to use way less, so the other condition would apply.


anaprietonem:
Example plots o96 - https://mlflow.ecmwf.int/#/experiments/5/runs/000304c17e304ac9bc353a7d0931d884/artifacts
Example plots n320 - https://mlflow.ecmwf.int/#/experiments/5/runs/f13f771ed2ac43268a75471093d71889

sahahner (approved):
Looks good to me.

HCookie (dismissed):

mchantry:
Has this been tested at a few resolutions?

anaprietonem:
> Has this been tested at a few resolutions?

Worked with @JPXKQX on this and should be now working fine for LAM

HCookie (approved):

PR #153, opened 2024-11-20
full shuffle of the dataset

ssmmnn11:
We currently split dates between tasks / workers and then the shuffling is performed in these subsets. These changes shuffle the dataset consistently on all workers and then each worker takes its subset from the shuffled indices. 

Thx to @JoffreyDumontLeBrazidec for testing.


theissenhelen (approved):

mchantry:
This can be merged.

PR #154, opened 2024-11-21
Validation and Training dataset dates assertion

gabrieloks:
Currently, the code uses an assertion to ensure that the validation dataset's start date occurs after the training dataset's end date. However, this check should issue a warning instead of being a blocking assertion.

### DIFF
--- a/src/anemoi/training/data/datamodule.py
+++ b/src/anemoi/training/data/datamodule.py
@@ -142,10 +142,12 @@ class AnemoiDatasetsDataModule(pl.LightningDataModule):
     def ds_valid(self) -> NativeGridDataset:
         r = max(self.rollout, self.config.dataloader.get("validation_rollout", 1))
 
-        assert self.config.dataloader.training.end < self.config.dataloader.validation.start, (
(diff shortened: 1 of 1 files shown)


src/anemoi/training/data/datamodule.py
```diff
@@ -142,10 +142,12 @@ def ds_train(self) -> NativeGridDataset:
     def ds_valid(self) -> NativeGridDataset:
         r = max(self.rollout, self.config.dataloader.get("validation_rollout", 1))
 
-        assert self.config.dataloader.training.end < self.config.dataloader.validation.start, (
-            f"Training end date {self.config.dataloader.training.end} is not before"
-            f"validation start date {self.config.dataloader.validation.start}"
-        )
+        if not self.config.dataloader.training.end < self.config.dataloader.validation.start:
+            LOGGER.info(
```
HCookie, line 146:
```suggestion
            LOGGER.warning(
```
I feel as if a warning would be more appropriate.

HCookie:
When you merge please refine the commit message to be more informative, and remove the superfolous commits

PR #160, opened 2024-11-25
hotfix: Expand scalar to prevent index out of bound error

HCookie:
Fix issue with index out of bounds when using `scalar_indices` with a `scalar` without a variable dimension


jswijnands (approved):
I have tested this and it now allows for the use of a limited area mask (i.e., a `scalar` without a variable dimension)

PR #165, opened 2024-11-25
Callback PlotHistogram breaks if only one variable is specified.

### LINKED ISSUE #163: Callback PlotHistogram breaks if only one variable is specified.
### What happened?

Instead of creating the plot for just one variable, it breaks when trying to iterate through the subplots.

Just list one parameter for the histogram plot

### Version

develop

atos/leonardo

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

_No response_

sahahner:
Fix #163 

### What happened?

Instead of creating the plot for just one variable, it breaks when trying to iterate through the subplots.

List only one parameter for the histogram or spectrum plot

### Solution

If only one variable is listed, create a list with only one entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ Keep it human-readable, your future self will thank you!
 ## [Unreleased](https://github.com/ecmwf/anemoi-training/compare/0.3.0...HEAD)
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


HCookie (approved):

gabrieloks (approved):
Great fix, thanks!

PR #168, opened 2024-11-27
Feature: Temporal interpolation

Magnus-SI:
Adds temporal interpolation functionality to anemoi. The idea is that a 6 or 12 hour forecaster might yield better predictions going days out than a 1 hour forecaster, as it has to make fewer auto-regressive steps. To produce the hourly predictions still, we can use the information available from the forecaster, e.g. hours 12 and 18 as input to predict hours 13-17. These predictions are made individually, assisted by some information about the target time as input.

This is a work in progress, parts of the implementation can be found on the corresponding branch of anemoi-models.

**Implemented**
- A prototype that runs (with decreasing loss, but haven't done any full scale training to verify yet). The interpolator itself is implemented as GraphInterpolator in interpolator.py, with train.py generalized to call a config-based model. If nothing is specified, it defaults to GraphForecaster.
- The config options multistep and rollout are not applicable to the interpolation case, so I added the option to explicitly state which time steps to use as input, and as a target for the model. This also enables non-regular input for the forecaster, e.g. using input at 0, -1, and -6 hours to make a 1 hour forecast.
- A corresponding change in the valid dates function, which also includes the possibility for data to be extracted across missing dates if the requested indices can cross the gap. E.g. with a missing date at index 8, and requested dates at 0,3,6. Indices 3,4 and 6,7 are still usable, which would not have been represented correctly before.
- Option to add temporal forcings at the target time as an input to the model, this is necessary for interpolation at multiple distinct hours with the same model weights. 

**To do**
- Add support for custom forcing parameters beyond those in the dataset. As of now, only the target time as a fractional difference between the input times is possible and used by default. I will instead move this to a config instantiated object.
- Run a full scale training and compare with previous results (from the old aifs-mono where I first implemented this)
- Also do a short training with the forecaster and see that the results are as expected.
- Implement a way to avoid crossing model runs. When training on data with analysis less frequent than the target interpolation frequency, one has to switch between different model runs at certain points in the dataset. If we for instance use 18 continuous hours of a run before switching, the change from hour 29 of one run to hour 12 of the next run will not represent a physical change in the state of the atmosphere. To avoid training on this, we thus need to consider all sets of inputs and targets that will cross this gap as invalid. I will add this to the usable_indices function.

**Questions**
- In train.py I could not get instantiate to work as an instance of the forecaster/interpolator because "config" is a kwarg of instantiate itself, and thus cannot be included in the additional kwargs. Instead I used a combination of importlib and getattr, but if there is a way to use instantiate here, I can replace it.
- Does anemoi datasets not support extracting dates with irregular indices? I could not get open_dataset to work with indices that do not follow a regular sliceable pattern. Either with an array or a tuple/list of indices. Is there a way around this?
Although a simple interpolation setup like using hours 0 and 6 to predict hours 1-5 yields a regular range from 0 to 6, irregular ranges would enable more complex setups for both the forecaster and interpolator.

### DIFF
--- a/src/anemoi/training/train/interpolator.py
+++ b/src/anemoi/training/train/interpolator.py
@@ -1,0 +1,118 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+import logging
+import math
+import os
+from collections import defaultdict
+from collections.abc import Generator
+from collections.abc import Mapping
+from typing import Optional
+from typing import Union
+from operator import itemgetter
+
+import numpy as np
+import pytorch_lightning as pl
+import torch
+from anemoi.models.data_indices.collection import IndexCollection
+from anemoi.models.interface import AnemoiModelInterface
+from anemoi.utils.config import DotDict
+from hydra.utils import instantiate
+from omegaconf import DictConfig
+from omegaconf import OmegaConf
+from timm.scheduler import CosineLRScheduler
+from torch.distributed.distributed_c10d import ProcessGroup
+from torch.distributed.optim import ZeroRedundancyOptimizer
+from torch.utils.checkpoint import checkpoint
+from torch_geometric.data import HeteroData
+
+from anemoi.training.losses.utils import grad_scaler
+from anemoi.training.losses.weightedloss import BaseWeightedLoss
+from anemoi.training.utils.jsonify import map_config_to_primitives
+from anemoi.training.utils.masks import Boolean1DMask
+from anemoi.training.utils.masks import NoOutputMask
+
+from anemoi.training.train.forecaster import GraphForecaster
+
+LOGGER = logging.getLogger(__name__)
+
+class GraphInterpolator(GraphForecaster):
+    """Graph neural network interpolator for PyTorch Lightning."""
+
+    def __init__(
+        self,
+        *,
+        config: DictConfig,
+        graph_data: HeteroData,
+        statistics: dict,
+        data_indices: IndexCollection,
+        metadata: dict,
+    ) -> None:
+        """Initialize graph neural network interpolator.
+
+        Parameters
+        ----------
+        config : DictConfig
+            Job configuration
+        graph_data : HeteroData
+            Graph object
+        statistics : dict
+            Statistics of the training data
+        data_indices : IndexCollection
+            Indices of the training data,
+        metadata : dict
+            Provenance information
+
+        """
+        super().__init__(config = config, graph_data = graph_data, statistics = statistics, data_indices = data_indices, metadata = metadata)
+        self.target_forcing_indices = itemgetter(*config.training.target_forcing.data)(data_indices.data.input.name_to_index)
+        self.boundary_times = config.training.explicit_times.input
+        self.interp_times = config.training.explicit_times.target
+
+
+    def _step(
+        self,
+        batch: torch.Tensor,
+        batch_idx: int,
+        validation_mode: bool = False,
+    ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor]]:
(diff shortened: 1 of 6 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=168) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-training?pullRequest=168) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-training?pullRequest=168) it.</sub>

src/anemoi/training/train/train.py
```diff
@@ -144,10 +144,13 @@ def model(self) -> GraphForecaster:
...
         }
+        train_module = importlib.import_module(getattr(self.config.training, "train_module", "anemoi.training.train.forecaster"))
+        train_func = getattr(train_module, getattr(self.config.training, "train_function", "GraphForecaster"))
+        #NOTE: instantiate would be preferable, but I run into issues with "config" being the first kwarg of instantiate itself.
         if self.load_weights_only:
             LOGGER.info("Restoring only model weights from %s", self.last_checkpoint)
-            return GraphForecaster.load_from_checkpoint(self.last_checkpoint, **kwargs)
-        return GraphForecaster(**kwargs)
+            return train_func.load_from_checkpoint(self.last_checkpoint, **kwargs)
+        return train_func(**kwargs)
```
HCookie, line 154: I agree that the instantiate would be preferable. If we were to delay the instantiatation of the `model` within the Forecaster, it may be possible to mimic a hydra instantiate call.

The delay will be neccessary to support loading weights only

```
model = instantiate({'_target_':self.config.get('forecaster'), **kwargs)
if self.load_weights_only:
            LOGGER.info("Restoring only model weights from %s", self.last_checkpoint)
            return train_func.load_from_checkpoint(self.last_checkpoint, **kwargs)
return model
```

src/anemoi/training/train/interpolator.py
```diff
@@ -0,0 +1,118 @@
...
+from anemoi.training.utils.masks import Boolean1DMask
+from anemoi.training.utils.masks import NoOutputMask
+
+from anemoi.training.train.forecaster import GraphForecaster
+
+LOGGER = logging.getLogger(__name__)
+
+class GraphInterpolator(GraphForecaster):
+    """Graph neural network interpolator for PyTorch Lightning."""
+
```
HCookie, line 48: I like this work on the `Interpolator`. It's a good example that the `GraphForecaster` class needs some work and to be broken into a proper class structure.
What are your thoughts on which components are reusable and then in counter, which parts are typical to override?

Magnus-SI: There's a mix of both, as well as some components that are needed only for the forecaster and some only for the interpolator.

**Reusable**
- All of the __init__ function, except for rollout and multistep.
- All of the instantiable objects: loss, metrics, the model, etc.
- The scheduler and optimizers, which should maybe become an instantiated object anyway.
- The _training/validation_step_ functions
- _calculate_val_metrics_: by reusing the rollout_step label as interp_step instead.

**Overwritten**
- __step_ and _forward_

**Only for the forecaster/interpolator**
- _advance_input_ and _rollout_step_
- target forcings (although these could also be useful for the forecaster)


To avoid inheriting unused components with the Interpolator, we could consider using a framework class containing only the common components between the forecaster and interpolator, then have both inherit this class. However, that might be a bit too much when there are only two options thus far.
In fact, the forecaster can be seen as a special case of the interpolator, since the boundary can be specified as the multistep input, and the target can be any time, including the future. If I implement rollout functionality to the interpolator and make the target forcings optional, I think it should be able to do anything the forecaster can.

In my opinion, it would be the best approach to merge the two this way. It also enables the option to train a combined forecaster/interpolator, instead of having two separate models.
Do you agree with merging the two, or should I make a base framework class for both to inherit, or just keep them as is?

HCookie: I think I would lean towards making a base framework class. There are other use cases coming down the pipeline that would need this.
Although I am intrigued by the idea of have a class that can do both together.

PR #174, opened 2024-11-28
Store numpy arrays in checkpoints

b8raoult:
We add to the checkpoint the latitudes, longitudes, and any other supporting bumpy arrays (e.g. masks)

### DIFF
--- a/src/anemoi/training/diagnostics/callbacks/checkpoint.py
+++ b/src/anemoi/training/diagnostics/callbacks/checkpoint.py
@@ -149,16 +149,22 @@ class AnemoiCheckpoint(ModelCheckpoint):
             tmp_metadata = model.metadata
             model.metadata = None
 
-            metadata = dict(**tmp_metadata)
+            tmp_supporting_arrays = model.supporting_arrays
+            model.supporting_arrays = None
+
+            # Make sure we don't accidentally modidy these
+            metadata = tmp_metadata.copy()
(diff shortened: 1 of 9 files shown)


JPXKQX:
@b8raoult it seems that some tests need to be updated

src/anemoi/training/diagnostics/callbacks/checkpoint.py
```diff
@@ -149,16 +149,22 @@ def _save_checkpoint(self, trainer: pl.Trainer, lightning_checkpoint_filepath: s
             tmp_metadata = model.metadata
             model.metadata = None
 
-            metadata = dict(**tmp_metadata)
+            tmp_supporting_arrays = model.supporting_arrays
```
anaprietonem, line 152: If `supporting_arrays` is also a dictionary couldn’t we just extend the metadata dict and write it there? Just thinking that doing so would remove the need for all this code, as it will be part of the metadata. Not sure if this has implications with the `anemoi-inference` interface or other packages so just a suggestion!

b8raoult: No, because Numpy arrays are not JSON serialisable.

anaprietonem (dismissed):
Thanks Baudouin for addressing my previous comment, the PR LGTM!

dietervdb-meteo (approved):
Code looks good. Passed various tests.

PR #180, opened 2024-12-04
[PROFILER] dont crash if an env var isnt found

cathalobrien:
Fall back to an empty string if env vars arent found when writing the profiler report. Otherwise you would get an error right at the end of your job


anaprietonem (approved):
Thanks Cathal and Jan for spotting this, LGTM!

PR #181, opened 2024-12-04
fix: allow None as graph save_path setting.

fprill:
This small changeset fixes the `save_path` option:
- the `anemoi-graphs` package accepts `None`, see https://github.com/ecmwf/anemoi-graphs/blob/fc614171f45e757b14bac92188c4d0b785e41e46/src/anemoi/graphs/create.py#L183 
- the `anemoi-training` just passes a `None` setting from the hydra YAML now.


mchantry (approved):
Great contribution, thanks!

PR #186, opened 2024-12-05
Feature/tendency scaler

pinnstorm:
Draft PR for tendency-scaler following PressureLevelScaler example

### DIFF
--- a/src/anemoi/training/data/scaling.py
+++ b/src/anemoi/training/data/scaling.py
@@ -77,3 +77,28 @@ class NoPressureLevelScaler(BasePressureLevelScaler):
         del plev  # unused
         # no scaling, always return 1.0
         return 1.0
+
+
+class BasePrognosticScaler(ABC):
+    """Configurable method to scale prognostic variables based on data statistics and statistics_tendencies."""
+
+    @abstractmethod
+    def scaler(self, variable_stdev: float, variable_tendency_stdev: float) -> float: ...
+
+
+class NoPrognosticScaler(BasePrognosticScaler):
+    """Constant scaling by 1.0."""
+
+    @staticmethod
+    def scaler(variable_stdev: float, variable_tendency_stdev: float) -> float:
+        del variable_stdev, variable_tendency_stdev  # unused
+        # no scaling, always return 1.0
+        return 1.0
+
+
+class TendencyPrognosticScaler(BasePrognosticScaler):
+    """Constant scaling by 1.0."""
+
+    @staticmethod
+    def scaler(variable_stdev: float, variable_tendency_stdev: float) -> float:
+        return variable_stdev / variable_tendency_stdev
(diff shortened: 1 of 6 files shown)


src/anemoi/training/train/forecaster.py
```diff
@@ -293,12 +295,28 @@ def get_variable_scaling(
...
+                )
+                scaling = tendency_scaler.scaler(variable_stdev, variable_tendency_stdev)
+                LOGGER.debug("Parameter %(key)s is being scaled by statistic_tendencies by %(scaling).2f")
+                variable_loss_scaling[idx] *= scaling
             split = key.split("_")
             if len(split) > 1 and split[-1].isdigit():
                 # Apply pressure level scaling
                 if split[0] in config.training.variable_loss_scaling.pl:
-                    variable_loss_scaling[idx] = config.training.variable_loss_scaling.pl[
+                    variable_loss_scaling[idx] *= config.training.variable_loss_scaling.pl[
```
mc4117, line 319: I'm wondering if with the tendency scaling, the variable scaling factor set by the user should be ignored?

src/anemoi/training/data/scaling.py
```diff
@@ -77,3 +77,28 @@ def scaler(plev: float) -> np.ndarray:
...
+
+
+class BaseTendencyScaler(ABC):
+    """Configurable method to scale prognostic variables based on data statistics and statistics_tendencies."""
+
+    @abstractmethod
+    def scaler(self, variable_stdev: float, variable_tendency_stdev: float) -> float: ...
+
+
+class NoTendencyScaler(BaseTendencyScaler):
```
mc4117, line 89: I'm guessing this should be added to the config as the default?

src/anemoi/training/train/forecaster.py
```diff
@@ -293,12 +295,28 @@ def get_variable_scaling(
...
+
         for key, idx in data_indices.internal_model.output.name_to_index.items():
+            if idx in self.data_indices.internal_model.output.prognostic:
+                prog_idx = self.data_indices.data.output.name_to_index[key]
+                variable_stdev = 1 / self.model.pre_processors.processors.normalizer._norm_mul[prog_idx]
+                variable_tendency_stdev = (
+                    self.statistics_tendencies["stdev"][prog_idx] if self.statistics_tendencies else 1
+                )
+                scaling = tendency_scaler.scaler(variable_stdev, variable_tendency_stdev)
+                LOGGER.debug("Parameter %(key)s is being scaled by statistic_tendencies by %(scaling).2f")
```
mc4117, line 313: This didn't work for me but the following did:

LOGGER.debug(f"Parameter {key} is being scaled by statistic_tendencies by {scaling:.2f}")

mc4117:
I have tested this with the ensemble branch 

This is with the tendencies and without the pressure level scalings
image

and this is with the tendencies and the pressure level scalings (also down-weighted w)

image

src/anemoi/training/config/training/default.yaml
```diff
@@ -117,3 +117,7 @@ pressure_level_scaler:
   _target_: anemoi.training.data.scaling.ReluPressureLevelScaler
   minimum: 0.2
   slope: 0.001
+
```
mchantry, line 120: I would have a comment explaining briefly what tendency scaler does.

Issue #57, opened 2024-09-16
Training checkpoints (of large models) are unreadble due to metadata 


anaprietonem:
### What happened?

Training checkpoints for large models (num_channels equal or greater than 912) become unreadable by pytorch and hence can't be used to resume or fork runs. 

Note - for now this issue can be overcome by:
1. patching an existing checkpoint: zip -d model.ckpt archive/anemoi-metadata/ai-models.json
2. commenting out the line that writes the metadata to the training config: https://github.com/ecmwf/anemoi-training/blob/91d8c6e9a8ac275ded43f781aac031ccca36499d/src/anemoi/training/diagnostics/callbacks/__init__.py#L888C1-L889C1 

These are temporary solutions, since if we also reach a point where our inference checkpoints are that large then we would also need a fix to be able to run inference

### What are the steps to reproduce the bug?

1. Update the config to use a model with num_channels equal or greater than 912. 
2. Run anemoi-training train --config-name filename
3. Once finished, try to resume or fork the above run with `run_id: previous_run_id` or `fork_run_id: previous_run_id`
4. Code crashes when loading the checkpoint with following error: `RuntimeError: PytorchStreamReader failed reading zip archive: invalid header or archive is corrupted`

### Version

0.1.0

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell
Error executing job with overrides: []
Traceback (most recent call last):
  File "/etc/ecmwf/nfs/dh2_home_a/ecm1715/anemoi-training/src/anemoi/training/train/train.py", line 355, in main
    AnemoiTrainer(config).train()
  File "/etc/ecmwf/nfs/dh2_home_a/ecm1715/anemoi-training/src/anemoi/training/train/train.py", line 341, in train
    trainer.fit(
  File "/perm/ecm1715/conda/envs/aifs-dev/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 543, in fit
    call._call_and_handle_interrupt(
  File "/perm/ecm1715/conda/envs/aifs-dev/lib/python3.10/site-packages/pytorch_lightning/trainer/call.py", line 43, in _call_and_handle_interrupt
    return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
  File "/perm/ecm1715/conda/envs/aifs-dev/lib/python3.10/site-packages/pytorch_lightning/strategies/launchers/subprocess_script.py", line 105, in launch
    return function(*args, **kwargs)
[...]

PR #190, opened 2024-12-06
fix: remove saving of metadata for training ckpt

pinnstorm:
This PR would remove the saving of the metadata from the training checkpoint (it is still present in the inference ckpt). I am not sure if there is a reason we need it in both checkpoints?

This PR came out of the AIFSv03 training where we had quite a bit of pain restarting runs with the zip bug for checkpoints over a certain size. We had to use the work around `zip -d last.ckpt archive/anemoi-metadata/ai-models.json` many times, which this PR would get around. Very happy to be told we need this metadata in the training checkpoints! But if not this would avoid quite a bit of pain when resuming runs. 🙏


mchantry:
Fixes #57

mchantry (approved):
Thanks for this.

mchantry:
@pinnstorm can you put something in the changelog please

mchantry (approved):
Great, thanks Ewan!

PR #196, opened 2024-12-11
Fix mlflow authenticate bug when forking run with multiple nodes

havardhhaugen:
Issue:
see issue #195 

Fix:
Add a check for rank zero only when authenticating etc for a forked run.


anaprietonem:
Hey @havardhhaugen  thanks for opening this fix! does this just happen with forking or also if you try to resume the run? The proposed fix might break mlflow functionality related to the forking and resuming of runs when migrating between 2 servers. I will try to test it to get a better idea. But I think we should try to understand why the authenticate is failing with those large jobs. Maybe it's matter of the number of server requests. @gmertes  Do you have a view regarding that?

havardhhaugen:
It only happens when forking.I have not done any testing on the fix other than seeing that it now works for the specific job that crashed earlier. So it would be great if you could do some more thorough testing. My suspicion is also that it crashes due to the large number of requests without the rank zero statement, but as far as I can tell the error message from mlflow is empty so it's hard to tell.

PR #202, opened 2024-12-13
Fix 'all' validation metrics

HCookie:
# Improve Validation Metrics

Allow `all` to normalised and scaled when calculating the validation metrics.

Adds a config option to control which validation_metrics stay in normalised space and which scalars are applied.

```yaml
# List of validation metrics to keep in normalised space, and scalars to apply
# Cannot be used for metrics associated with variables that are remapped between the
# real space and internal model space.
scale_validation_metrics:
  scalars_to_apply: ['variable']
  metrics:
    - 'all'
```

## Testing
- [x] Tested in default running [here](https://mlflow.ecmwf.int/#/experiments/93/runs/2c9173ff907c4ec2baa9afc1458af5ab/model-metrics)
- [ ] Stretched Grid @jswijnands [here](https://mlflow.ecmwf.int/#/experiments/30/runs/654966120d9c4e618b71d1d4b5d89e88/model-metrics)
- [x] Remapped data @sahahner 

----
📚 Documentation preview 📚: https://anemoi-training--202.org.readthedocs.build/en/202/

### DIFF
--- a/src/anemoi/training/config/training/default.yaml
+++ b/src/anemoi/training/config/training/default.yaml
@@ -58,12 +58,16 @@ loss_gradient_scaling: False
 
 # Validation metrics calculation,
 # This may be a list, in which case all metrics will be calculated
-# and logged according to their name
+# and logged according to their name.
+# These metrics are calculated in the output model space, and thus
+# have undergone postprocessing.
 validation_metrics:
   # loss class to initialise
   - _target_: anemoi.training.losses.mse.WeightedMSELoss
     # Scalars to include in loss calculation
     # Cannot scale over the variable dimension due to possible remappings.
+    # Available scalars include:
+    # - 'loss_weights_mask': Giving imputed NaNs a zero weight in the loss function
     # Use the `scale_validation_metrics` section to variable scale.
     scalars: []
     # other kwargs
@@ -71,6 +75,8 @@ validation_metrics:
 
 # List of validation metrics to keep in normalised space, and scalars to be applied
 # Use '*' in reference all metrics, or a list of metric names.
+# Unlike above, variable scaling is possible due to these metrics being
+# calculated in the same way as the training loss, within the internal model space.
 scale_validation_metrics:
   scalars_to_apply: ['variable']
   metrics:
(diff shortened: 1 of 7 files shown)


src/anemoi/training/train/forecaster.py
```diff
@@ -251,6 +251,8 @@ def training_weights_for_imputed_variables(
                     loss_weights_mask = pre_processor.transform_loss_mask(loss_weights_mask)
             # update scaler with loss_weights_mask retrieved from preprocessors
             self.loss.update_scalar(scalar=loss_weights_mask.cpu(), name="loss_weights_mask")
+
+        self.scalars["loss_weights_mask"] = ((-2, -1), loss_weights_mask.cpu())
```
sahahner, line 255: This breaks if loss_weights_mask is not mentioned in the training-config file (which is the default). `loss_weights_mask` has to be defined before the if-statement.

src/anemoi/training/config/training/default.yaml
```diff
@@ -68,6 +68,15 @@ validation_metrics:
     # other kwargs
     ignore_nans: True
 
+# List of validation metrics to keep in normalised space, and scalars to apply
+# Cannot be used for metrics associated with variables that are remapped between the
+# real space and internal model space.
+scale_validation_metrics:
+  scalars_to_apply: ['variable']
+  metrics:
+    - 'all'
```
sahahner, line 77: If adding 'all' here is the new default, the runs break when applying the remapper since the number of variables does not match the number of scalars. Do we need an assert/warning that prevents from calling the weighted validation loss in the case the remapper is applied?

HCookie: Addressed in [f51fbb1](https://github.com/ecmwf/anemoi-training/pull/202/commits/f51fbb13ecd13f80aa5477f503a1c42d5aef5846)

src/anemoi/training/train/forecaster.py
HCookie: Fixed in [4f322fa](https://github.com/ecmwf/anemoi-training/pull/202/commits/4f322fa12e90df6f3b6f6666a9b50033ee841d97)

sahahner (commented):
This has been tested for the model training (no stretched grid approach) with and without using the remapper. The behavior is as expected. Looks good to me.

JesperDramsch (approved):
Looks like @sahahner gave it a good review and I can't see any reason not to approve.

Issue #203, opened 2024-12-14
Bug: check for stretched grid breaks running with minimal graph config


dietervdb-meteo:
### What happened?

The merging of [PR199](https://github.com/ecmwf/anemoi-training/pull/199/files) introduced the line
`if "lam_resolution" in getattr(config.graph.nodes.hidden, "node_builder", []):` in forecaster.py
This will result in an error in case one sets up a config where the graph part is the minimal
```yaml
graph:
  overwrite: False 
  data: data
  hidden: hidden
```
which is sufficient to load a graph previously stored on disk. Rather than trying to determine if the model is stretched grid or not from the graph part of the training config this should preferably be checked on the graph directly.

### What are the steps to reproduce the bug?

Run training with a config for which the graph part is as above.

### Version

current develop

### Platform (OS and architecture)

Lumi [irrelevant]

### Organisation

rmi

dietervdb-meteo:
@jswijnands , any simple fixes/ways around for this?

dietervdb-meteo:
As a temporary fix, the following minimal graph config still works [when running from a graph on disk]:
```yaml
graph:
  overwrite: False #This to take graph from file
  data: data #This and next line needed by AnemoiModelEncProcDec in anemoi-models
  hidden: hidden 
  # below needed to circumvent the bug reported here [it doesn't do anything else in the code]
  nodes:
    hidden:
      node_builder:
        _target_: anemoi.graphs.nodes.LimitedAreaTriNodes # options: ZarrDatasetNodes, NPZFileNodes, TriNodes
```

dietervdb-meteo:
As mentioned in https://github.com/ecmwf/anemoi-models/issues/92 , it would be elegant if the graph part of the training config when one wants to run from a graph on disk could be as minimal as:
```yaml
graph:
   overwrite: False #or even something like create: False
```

jswijnands:
Hi Dieter, I see. We have indeed discussed various ways on how to identify whether a model is a stretched grid or not. Your minimal graph config use case was not envisioned then. The temporary fix you suggested indeed works.

jswijnands:
The if statement below may solve this problem, so you can still use a minimal script:
```
if "nodes" in getattr(config, "graph", []) and
    "hidden" in getattr(config.graph, "nodes", []) and
    "node_builder" in getattr(config.graph.nodes, "hidden", []) and
    "lam_resolution" in getattr(config.graph.nodes.hidden, "node_builder", []):
```

[...]

PR #204, opened 2024-12-16
Fix/stretched grid check

jswijnands:
In the forecaster, there is a need to identify if we are dealing with a stretched grid model. The improvement in this PR also works if an existing graph is loaded from disk, in combination with a minimal config file specification of the graph.

This code update has been checked for stretched grid models.

Review suggestion (RMI): check if this change also works for LAM models when loading an existing graph from disk using a minimal configuration script.


dietervdb-meteo (approved):
Code looks good. Tested on  training of lam model with minimal config and graph from disk. This closes #203  .

PR #208, opened 2024-12-19
Exclude nans from increment and persistent error color bars

anaprietonem:
Fix for https://github.com/ecmwf/anemoi-core/issues/12

[WIP - pending feedback from @lzampier ]


lzampier:
The plots on @anaprietonem test training are correct with this fix.

